PR #36, opened 2018-05-28
🚑(minishift) fix wrong variable name

cchaudier:
Bug : with VirtualBox the 4GB memory opt was erased instead of be overridden.


sampaccoud (approved):
Could comment the criticality of the bug and the choice of emoji 😉 but GTM anyway.

PR #38, opened 2018-05-28
Use an environment variable to configure checking SSL in openshift_raw module

sampaccoud:
## Purpose

The openshift_raw module can be configured via an environment variable to check SSL certificates before connecting to an OpenShift instance or not. This avoids having to set it on each object in our playbooks.

## Proposal

-  [x] Add K8S_AUTH_VERIFY_SSL=no to our default environment file
- [x] Remove all verify_ssl parameters in playbooks


jmaupetit (approved):
:shipit:

PR #42, opened 2018-05-29
Customize settings for patient0 demo

sampaccoud:
Validate that modifying settings via the OpenShift template is effective by customizing the demo.


jmaupetit (approved):
:shipit:

PR #53, opened 2018-06-07
🎨(app) make nginx an app-specific service

jmaupetit:
## Purpose

Until now, we considered that nginx may be a single service used by many apps at the same time. Even if this pattern is quite reasonable, it offers less configurability if we want to deploy not all but a selection of applications Arnold can play with. Hence we decided to split the monolithic nginx service into multiple nginx services (one per application). 

## Proposal

- [x] split the common/nginx svc
- [x] split the common/nginx dc
- [x] split nginx configMaps
- [x] redefine app-specific routes to point to the appropriate nginx svc
- [x] ~improve route aliases definition to be app specific~
- [x] add a `redirect` app whose responsibility is to handle route aliases
- [x] move ES common service to richie


group_vars/all/main.yml
```diff
@@ -82,11 +82,12 @@ richie_database:
 
 # Openshift template files
 openshift_deployments:
+  - openshift/common/dc/elasticsearch.yml
```
sampaccoud: In fact it's not common because richie and edxapp require different versions of ES. So this one is dedicated to Richie.

templates/openshift/common/dc/redirect.yml.j2
```diff
@@ -0,0 +1,33 @@
+apiVersion: v1
```
sampaccoud: I was expecting a "redirect" directory at the same level as "edxapp" and "richie" but the svc/dc would still be nginx:

 templates/openshift/redirect/dc/nginx.yml.j2
 templates/openshift/redirect/svc/nginx.yml.j2

WDYT?

group_vars/all/main.yml
jmaupetit: True. For now, I will make this dc/svc richie-specific. We will add ES for edxapp later.

templates/openshift/common/dc/redirect.yml.j2
jmaupetit: Totally right. Let's do this.

group_vars/all/main.yml
cchaudier: I think we have to abandon the `common` directory.
I think we should consider the `templates/openshift` directory as a collection of usable micro-services for an application.
I propose that we adopt the following naming convention:
`templates/openshift/<micro-service-name>/{svc,dc,pcv,...}`

If we do like this in the future we can find all templates use by a microservice and not hardcode the path in an ansible vars.

templates/openshift/common/dc/redirect.yml.j2
cchaudier: It's a very good idea and close to the naming convention I propose.

group_vars/all/main.yml
jmaupetit: Agreed! I'll switch to this pattern next week.

templates/openshift/edxapp/dc/nginx.yml.j2
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    app: nginx
```
cchaudier, line 5: All version of an app are grouped on th same app on OpenShift ?
Why do not include the deployment_stamp also ?
The app must be `edxapp` no ?

templates/openshift/common/svc/redirect.yml.j2
```diff
@@ -0,0 +1,19 @@
...
+kind: Service
+metadata:
+  labels:
+    app: redirect
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: redirect-{{ deployment_stamp }}
+  namespace: "{{ project_name }}"
+spec:
+  ports:
+{% for port in nginx_ports %}
```
cchaudier: If we split nigix into app micro-service whe don't need the nginx_ports and this loop.

cchaudier (changes requested):
Some question and proposition.
Great job !

templates/openshift/common/svc/redirect.yml.j2
jmaupetit: Yes. I'll do this refactoring next week.

templates/openshift/edxapp/dc/nginx.yml.j2
jmaupetit: Yes! and that will be part of my next refactoring. IMO this is not part of the scope of this PR.

sampaccoud (approved):

templates/openshift/edxapp/dc/nginx.yml.j2
cchaudier: Ok

templates/openshift/common/svc/redirect.yml.j2
cchaudier: ok

cchaudier (approved):

PR #54, opened 2018-06-08
🎨(playbook) better separate init vs deploy scope

jmaupetit:
## Purpose

Feedback from our first Arnold's workshop showed us that the `init_project.yml` playbook scope was not clearly defined and should not make an initial deployment of the project. 
Hence we decided to separate the project initialization from deployments. This makes the deployment strategy more predictable/reproducible.

## Proposal

- [x] make the `init_project.yml` playbook initialize the project (it no longer makes the initial deployment)
- [x] remove unused `create_objects.yml` playbook


init_project.yml
```diff
@@ -1,6 +1,26 @@
...
+      debug: msg="Project name {{ project_name }} - domain {{ domain_name }}"
+
+    - name: Make sure base objects exist in OpenShift and are up-to-date
+      openshift_raw:
+        definition: "{{ lookup('template', 'templates/' + item + '.j2') | from_yaml }}"
+        state: present
+      with_items:
+        - "{{ openshift_volumes }}"
+        - "{{ openshift_endpoints }}"
+      when: item is not none
```
sampaccoud, line 24: I'm thinking: since routes are going to be fixed (previous, current, next) they should also be created in init. 
Could be part of @cchaudier's PR though WDYT?

sampaccoud (approved):
On minor suggestion but GTM. Wow this is much simpler ;-)

deploy.yml
```diff
@@ -1,7 +1,6 @@
 ---
 # This playbook deploys a whole new stack with the current `deployment_stamp`
 # (see main.yml)
-- import_playbook: create_config.yml
```
cchaudier, line 4: How to deploy the configs if they change ?

init_project.yml
```diff
@@ -1,6 +1,26 @@
 ---
 - import_playbook: create_project.yml
 - import_playbook: create_secrets.yml
-- import_playbook: create_config.yml
 - import_playbook: create_acme.yml
-- import_playbook: create_objects.yml
+
```
cchaudier: why don't import the deploy playbook on init to perform the initial deploy ?

deploy.yml
```diff
@@ -33,3 +32,5 @@
       - "{{ openshift_services }}"
       - "{{ openshift_jobs }}"
       - "{{ openshift_routes }}"
+
+    - import_tasks: tasks/create_route_aliases.yml
```
cchaudier: Whe can use the [Ansible tags](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html) to deploy the app, the routes or th configs like this : 
```bash
docker run --rm -it \	
    --env-file env.d/production \	
    arnold \	
    ansible-playbook deploy.yml -e "customer=myCustomer env_type=myEnv" \
         --tags "configuration,app"
```
Or : 
```bash
docker run --rm -it \	
    --env-file env.d/production \	
    arnold \	
    ansible-playbook deploy.yml -e "customer=myCustomer env_type=myEnv" \
         --tags "routes"
```

It can be also on a next issues.

cchaudier (commented):
Greate job.
I've somes question and some improvement for this PR or a new issue.

init_project.yml
cchaudier: I can do this yes.

deploy.yml
jmaupetit: The`tasks/create_config.yml` tasks are imported in init and deploy.

deploy.yml
jmaupetit: That's indeed a nice to have.

jmaupetit: Can you create a new issue to better describe what you have in mind?

init_project.yml
jmaupetit: Good idea!

PR #57, opened 2018-06-08
✨(job) ease job restart

jmaupetit:
## Purpose

At this time, as jobs are immutable, re-run a failed job is a painful task: one need to delete it using `oc` or the OpenShift console, and then re-create it _via_ the `create_object` playbook.

## Proposal

- [x] Add a unique `job_stamp` to jobs (short UUID)

This `job_stamp` is set during the deployment and when using the `create_object` playbook.

_nota bene_: I choose to keep using the create_object.yml playbook to restart jobs instead of creating a new one, as almost the entire playbook would have been duplicated.

### DIFF
--- a/create_object.yml
+++ b/create_object.yml
@@ -7,7 +7,8 @@
 # Usage:
 #
 #   $ bin/ansible-playbook create_object.yml \
-#       -e "object_template=templates/openshift/edxapp/job/import_demo_course.yml.j2"
+#       -e "object_template=templates/openshift/edxapp/job/import_demo_course.yml.j2" \
+#       -e "deployment_stamp=d-180611-08h46m30s"
 #
 
 - hosts: local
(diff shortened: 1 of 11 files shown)


create_object.yml
```diff
@@ -7,7 +7,8 @@
 # Usage:
 #
 #   $ bin/ansible-playbook create_object.yml \
-#       -e "object_template=templates/openshift/edxapp/job/import_demo_course.yml.j2"
+#       -e "object_template=templates/openshift/edxapp/job/import_demo_course.yml.j2" \
+#       -e "deployment_stamp=2018-06-07-15-41-31-838798865"
```
sampaccoud: I really think that we should get rid of the hash and that a stamp to the second is really all we need. It could be formatted like this for clarity:
180607-15h41m31s

Same for deployment stamps.

So we could prepend them with D for deployment and J for job : 

- D-180607-15h41m31s
- J-180607-15h41m33s

jmaupetit: The last part (_i.e._ `838798865`) represents the nanoseconds of the stamp date. That's a security to avoid duplicates, but you are right in 99.99% cases it will be fine without this. Let's try proposed patterns.

**Important remark**: those stamps should the shortest possible as they are concatenated with the object name (_e.g._ `richie-regenerate-indexes`) and the total size of this string should not be longer than 63 characters.

sampaccoud: yes for humans also it's better if it's as short as possible. I think you can make it 100% of the cases because a CI trying to deploy the same project twice in a second is probably misbehaving?

jmaupetit: Probably true. And in a blue/green deployment strategy, this "should not" impact the current route. But, you know, as a scientist, I am always doubting... :wink:

cchaudier: Can you add a sugar script (e.g. `bin/launch_job patient0 staging import_demo_course`) who automaticaly calculate the job_stamp ?

jmaupetit: The `job_stamp` is already handled by the `create_object` playbook. We cannot guess the `deployment_stamp` as, given a particular customer and environment, we need to know which stack we are targeting, _i.e._ previous, next or current?

But adding a sugar script to ease job restart is a pretty good idea!

jmaupetit: @cchaudier I've added this script in my last fixup. Check this out!

sampaccoud: don't forget to change the stamp format in this docstring

jmaupetit: Done

jmaupetit:
@cchaudier the `customer` and `env_type` are now automatically guessed given a `deployment_stamp` :sunglasses:

cchaudier (approved):
GTM ! :+1:

PR #58, opened 2018-06-10
Feature/fix bin dev

cchaudier:
## Bug

bin/dev fail after lint correction

## Fix

In bin/dev we build a command line.
So we must ignore the shellsheck SC2086 rule to split on spaces
The docker image dind used in gitlab-ci don't contain bash.              
We add bash because it used on shell scripts launch in the CI jobs.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -50,7 +50,8 @@ function _run-minishift() {
     fi
 
     echo "Starting minishift with options: ${opts}"
-    minishift start "${opts}"
+    # shellcheck disable=SC2086
+    minishift start ${opts}
     minishift status
 
     echo -e "\n**** MiniShift is up and running ****"


.gitlab-ci.yml
```diff
@@ -22,6 +22,7 @@ stages:
 
 # Before executing a job's script, we check the output of docker info
 before_script:
+  - apk add --update bash
```
jmaupetit: Why do we need to update bash?

.gitlab-ci.yml
```diff
@@ -30,6 +31,7 @@ build:image:
   script:
     - bin/build
     - mkdir -p docker/images
+    - ARNOLD_IMAGE="arnold:$(tr -d '\n' < VERSION)"
```
jmaupetit: Why not using YAML anchors to avoid repetition?

.gitlab-ci.yml
```diff
@@ -30,6 +31,7 @@ build:image:
...
     - docker save -o docker/images/arnold.tar ${ARNOLD_IMAGE}
   # Store the built as an artifact that can be reused in other jobs
   artifacts:
```
jmaupetit: BTW, I think these changes should not be part of this PR (out of scope).

bin/dev
```diff
@@ -50,7 +50,8 @@ function _run-minishift() {
     fi
 
     echo "Starting minishift with options: ${opts}"
-    minishift start "${opts}"
+    # shellcheck disable=SC2086
+    minishift start ${opts}
```
jmaupetit, line 54: :+1:

.gitlab-ci.yml
cchaudier: It's explain on the PR message, it's for use the build script.

.gitlab-ci.yml
cchaudier: No time for this for the moment but yes we must use anchors.
I'm open an issue for sync circleCI and gitlab-ci files.

.gitlab-ci.yml
cchaudier: You're right.
I'm open an issue for sync circleCI and gitlab-ci files.

.gitlab-ci.yml
jmaupetit: Can you add a comment instead?

cchaudier: I'm going to do this on the issue #59.
I've remove the commit for gitlab-ci.

jmaupetit (approved):
LGTM!

Issue #47, opened 2018-06-05
Switch routes for blue green deployments


sampaccoud:
## Purpose

Currently, our deployments create new complete stacks including routes to access each stack.
In a functional architecture, we need to manage routes differently to allow uninterrupted service when making a new deployment.

## Proposal

Stop creating routes together with the stack of svc/dc/ep.

Create only 3 routes for each service. The following hosts are examples for the lms:

- previous: "**previous**.lms.{{ project_name }}.{{ domain_name }}"
- current: "lms.{{ project_name }}.{{ domain_name }}"
- next: "**next**.lms.{{ project_name }}.{{ domain_name }}"

Each time we make a new deployment, the following actions are taken:

1) create the new stack,
2) create/update the next route:
    - if the "next" route pre-existed, update it to point to the new stack. Delete the stack that was previously pointed by the "next" route, making sure that it is not pointed by the "current" route (it is normally not possible but could happen if something went wrong in a route switch... so let's check that explicitly before deleting it),
    - if the "next" route did not exist, create it and point it to the new stack.

Each time we confirm a deployment by triggering a route switch, the following actions are taken:

1) point the "current" route to the stack that was pointed by the "next" route,
2) point the "previous" route to the stack that was pointed by the "current" route just before we changed it in 1)
3) delete the stack that was pointed by the "previous" route just before we changed it in 2) 
4) delete the next route

jmaupetit:
This is fixed since 7ee5e601acde595743045d40628415d0f0c90a94

Issue #44, opened 2018-05-30
Make projects customizable


sampaccoud:
## Purpose

We project to be customizable so that anyone can use Anrold as a tool to build their own project.

## Proposal

Add a new directory with the following structure:

```
├── examples
│   └── hello
│       └── templates
│           └── ...
│       └── group_vars
│           └── ...
│   └── openedx
│       └── templates
│           └── ...
│       └── group_vars
│           └── ...
│   └── richie
│       └── templates
│           └── ...
│       └── group_vars
│           └── ...

```

In order to deploy one of these projects, the scripts should run docker mounting the `templates` and `group_vars` directories in the *arnold directory.

- [ ] move the existing specific configuration to a private repository
- [ ] delete the `files` directory as it won't be used for the moment
- [ ] add the examples directory and split the current templates/vars into 3 projects: hello, openedx, richie
- [ ] modify the `_docker_run` bash command to take the project name as an argument.

jmaupetit:
This has been implemented in #62

Issue #40, opened 2018-05-28
Allow customizing the list of configmaps per customers and environments


sampaccoud:
## purpose

The playbook that creates configmaps automatically find them in the `templates/configmap` directory and pushes them all to OpenShift. We want to be able to define which configmaps are created for each customer or env_type. 

For example, the "edxapp-fixtures" configmap should only be created for the `development` or `feature` environments.

## Proposal

Apply the same mechanism as for OpenShift objects to configmaps:

- [ ] List configmaps in an Ansible variable called `openshift_configmaps` that can be overridden by customer/env_type,
- [ ] Stop autodiscovering configmaps in the playbook but use the `openshift_configmaps` variable.

jmaupetit:
Closing this as this feature already has already been implemented.

PR #64, opened 2018-06-13
feature/improve vars setting

cchaudier:
## Purpose

To prepare issues #40, #44 and #47 we must set the `env_type` and `customer` vars after the `apps` default vars.
The default app vars must be set on `include_vars` but it's after `play vars_files` in [Ansible Variable Precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable).

## Proposal

So we load `env_type` and `customers` vars on `include_vars` step, and in the future after the task who set apps vars.

### DIFF
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -1,0 +1,14 @@
+---
+# tasks/set_vars.yml
+
+# TODO : set vars for apps here
+
+- name: Include customer env_type vars
+  include_vars:
+    file: "{{ item }}"
+  with_items:
+    - "group_vars/customer/{{ customer }}.yml"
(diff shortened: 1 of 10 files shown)


tasks/set_vars.yml
```diff
@@ -0,0 +1,14 @@
+---
+# tasks/set_vars.yml
+
+# TODO : set vars of apps here
```
jmaupetit: set vars for apps

jmaupetit (approved):
Minor typo, else

tasks/set_vars.yml
cchaudier: done

PR #62, opened 2018-06-12
Move project tree to an application-driven logic

jmaupetit:
## Purpose

We need to make applications easily pluggable to use Arnold as a generic way to deploy them. This means that we need to move the project tree to a structure easily overridable when using Arnold's docker image. Moreover, we need to be able to easily deploy a single application without having to comment or update group vars definitions.

## Proposal

- [x] move OpenShift object templates to the `apps/` directory
- [x] move ConfigMaps to the target application configmaps directory (_e.g._  `apps/{{ app }}/configmap`)
- [x] ~move secrets to the target application secrets directory (_e.g._  `apps/{{ app }}/secret`)~ edit: postponed to a future PR
- [x] ~move application-specific settings to a file (_e.g._  `apps/{{ app }}/group_vars/main.yml`)~ edit: postponed to a future PR
- [x] create application objects by scanning _ad hoc_ directories instead of hard-coding them in a configuration (_i.e._ `apps/{{ app }}/dc/*.yml.j2` for DCs) \*
- [x] ~get rid of the `_common` templates?~
- [x] ~update the documentation~ edit: postponed to a future PR

\* we need to find a way to easily modify this mechanism for a particular environment, _e.g._ ignore endpoints in development.

This work is related to the following issues: #40, #44

PR #60, opened 2018-06-11
Switch routes for blue green deployments

cchaudier:
## Purpose

Currently, our deployments create new complete stacks including routes to access each stack.
In a functional architecture, we need to manage routes differently to allow uninterrupted service when making a new deployment.

## Proposal

Stop creating routes together with the stack of svc/dc/ep.

Create only 3 routes for each service. The following hosts are examples for the lms:

- previous: "**previous**.lms.{{ project_name }}.{{ domain_name }}"
- current: "lms.{{ project_name }}.{{ domain_name }}"
- next: "**next**.lms.{{ project_name }}.{{ domain_name }}"

Each time we make a new deployment, the following actions are taken:

1) create the new stack,
2) create/update the next route:
    - if the "next" route exists, update it to point to the new stack. 
              - making sure that it is not pointed by the "current" route (it is normally not possible but could happen if something went wrong in a route switch... so let's check that explicitly before deleting it),
              - Delete the stack that was previously pointed by the "next" route, 

    - if the "next" route does not exist, create it and point it to the new stack.

Each time we confirm a deployment by triggering a route switch, the following actions are taken:

1) point the "current" route to the stack that was pointed by the "next" route,
2) point the "previous" route to the stack that was pointed by the "current" route just before we changed it in 1)
3) delete the stack that was pointed by the "previous" route just before we changed it in 2) 
4) patch the next route to the init stamp

## TODO
~~- [x] Create fixed routes on init project~~ now on PR #66 
- [x] Update the next route after deployment
- [x] Delete the previous stack
- [x] Switch routes next to current
- [x] Switch routes current to previous

PR #66, opened 2018-06-14
Feature/create statics routes

cchaudier:
## Purpose

Init_project.yml must creates the 3 defaults routes for each service who need them.
Split of the PR #60, because the code is needed in the PR #62

## Proposal

Stop creating routes together with the stack of svc/dc/ep.

Create only 3 routes for each service. The following hosts are examples for the lms:

- previous: "**previous**.lms.{{ env_type }}.{{ domain_name }}"
- current: "lms.{{ env_type }}.{{ domain_name }}"
- next: "**next**.lms.{{ env_type }}.{{ domain_name }}"

On init create the routes who point to the service that does not yet exist but that will be created by the deployment.

## TODO
- [x] Create fixed routes on init project


group_vars/all/main.yml
```diff
@@ -3,6 +3,28 @@ env_type: development  # default to be overriden on command line
...
+#     services:
+#       - name:     richie
+#       - name:     richie-nginx
+#         host:     "{{ env_type }}.richie.{{ domain_name }}"
+#       - name:     postgresql
+#         endpoint: true  # default False
+apps:
+
+
+## _____ soon deprecated below _________________________________________________
```
```yaml
## /!\  Variables below will soon be deprecated /!\ ##
- foo:
```

init_project.yml
```diff
@@ -24,3 +24,9 @@
       when: item is not none
 
     - import_tasks: tasks/create_route_aliases.yml
+
+    - include_tasks: tasks/loop_foreach_apps.yml
+      vars:
+        tasks_to_launch:
+          - create_static_services_routes
+      tags: route
```
jmaupetit: I propose the following naming schema:

```yaml
   - include_tasks: tasks/run_tasks_for_apps.yml
     vars:
       tasks:
         - create_static_services_routes
     tags: route
```

tasks/loop_foreach_apps.yml
```diff
@@ -0,0 +1,16 @@
+---
+# This task loop to all apps in {{ apps }},
+# define the {{ app }} vars and
+# launch the helper task `loop_foreach_tasks_for_current_app.yml`
```
jmaupetit: WDYT about the following?

> This task loops over all applications in {{ apps }}, defines the {{ app }} var and runs the xxx task for this app.

tasks/loop_foreach_tasks_for_current_app.yml
```diff
@@ -0,0 +1,8 @@
+---
+# This task loop to all tasks in {{ tasks_to_launch }}
+# and launch it.
```
jmaupetit: > This task runs all tasks in {{ tasks }} for an app.

tasks/loop_foreach_tasks_for_current_app.yml
```diff
@@ -0,0 +1,8 @@
...
+
+- include_tasks: "{{ tsk_to_launch }}.yml"
+  loop: "{{ tasks_to_launch }}"
+  loop_control:
+    loop_var: tsk_to_launch
```
jmaupetit: Following new naming schema, this would be simpler, no?

```yaml
- include_tasks: "{{ task }}.yml"
  loop: "{{ tasks }}"
  loop_control:
    loop_var: task
```

jmaupetit (commented):
I mostly have naming changes to submit, else :muscle:

group_vars/all/main.yml
cchaudier: done with this for the 80 char 
```
## /!\  Variables below will soon be deprecated /!\ __________________________##
```

init_project.yml
cchaudier: done

tasks/loop_foreach_apps.yml
cchaudier: done

tasks/loop_foreach_tasks_for_current_app.yml
cchaudier: done

tasks/loop_foreach_tasks_for_current_app.yml
cchaudier: done

jmaupetit (approved):

PR #68, opened 2018-06-21
Allow application filtering to select apps to work on

jmaupetit:
## Purpose

When working with Arnold, we often want to deploy a single application (or a subset of all available applications). Since Arnold now auto-discovers applications and works with all of those apps, it wasn't possible to easily achieve app selection.

## Proposal

This work implements an application filter given an extra var passed to playbooks, _e.g._:

```bash
$ ansible-playbook foo.yml -e "filter_apps='richie,redirect'"
```

### DIFF
--- a/tasks/set_vars.yml
+++ b/tasks/set_vars.yml
@@ -4,7 +4,28 @@
 - name: Lookup applications and set the apps var
   set_fact: apps="{{ lookup('pipe', 'python plugins/utils/lookup_apps.py') | from_yaml }}"
 
-- name: Display looked up applications
+# If the playbook is invoqued with the apps_filter extra var, we restrict apps
+# to the list passed as an argument, e.g.:
+#
+#     $ ansible-playbook foo.yml -e "apps_filter='richie,edxapp'"
+#
+# This task will build the jmespath query to restrict apps to the selected
+# one(s), e.g.:
+#
+#    [?name=='richie' || name=='edxapp']
+- name: Set app filtering query
+  set_fact:
+    # We use escaped double quotes in the first regex_replace filter to be able
+    # to add single quotes around app name in the query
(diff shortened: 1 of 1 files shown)


tasks/set_vars.yml
```diff
@@ -4,7 +4,28 @@
 - name: Lookup applications and set the apps var
   set_fact: apps="{{ lookup('pipe', 'python plugins/utils/lookup_apps.py') | from_yaml }}"
 
-- name: Display looked up applications
+# If the playbook is invoqued with the filter_apps extra var, we restrict apps
```
sampaccoud: if we follow the same logic as for "customer" and "env_type", it should be called "app_filter" rather then "filter_apps". WDYT?

jmaupetit: Given the name of this branch, I think it makes sense :wink: The past me already agreed.

tasks/set_vars.yml
```diff
@@ -4,7 +4,28 @@
 - name: Lookup applications and set the apps var
   set_fact: apps="{{ lookup('pipe', 'python plugins/utils/lookup_apps.py') | from_yaml }}"
 
-- name: Display looked up applications
+# If the playbook is invoqued with the app_filter extra var, we restrict apps
+# to the list passed as an argument, e.g.:
+#
+#     $ ansible-playbook foo.yml -e "app_filter='richie,redirect'"
```
cchaudier: You don't think apps_filter is a better name for this vars ?

tasks/set_vars.yml
```diff
@@ -4,7 +4,28 @@
...
+#
+# This task will build the jmespath query to restrict apps to the selected
+# one(s), e.g.:
+#
+#    [?name=='richie' || name=='redirect']
```
cchaudier: For me redirect, like openshift-acme, must be present in all project.
Is not a good exemple, IMHO.

cchaudier (commented):
Yes good job.
I have some questions.

tasks/set_vars.yml
jmaupetit: Seems relevant.

cchaudier (approved):

PR #72, opened 2018-06-28
✨(playbook) add apps metadata from group_vars

jmaupetit:
## Purpose

We need a way to add metadata to applications and services listed in the `apps` variable.

## Proposal

We now allow to define an `apps` var for a customer (_e.g._ `groups_vars/customer/patient0.yml`) or an environment (_e.g._ `groups_vars/env_type/staging.yml`) to add metadata to applications and their services.

Known limitation: for now, we only support the merge strategy between discovered apps and groups_vars defined metadata. This can be refined later if there is a need to strictly override some definitions (and not only add metadata).


plugins/filter/merge.py
```diff
@@ -0,0 +1,45 @@
+"""
```
cchaudier, line 1: Can we rename the file `merge_with_app` ?

cchaudier (approved):
Great job !

plugins/filter/merge.py
jmaupetit: My plan was to add merge-related filters in this file, not only the `merge_with_app` filter.

PR #76, opened 2018-07-06
Refactor plugins

jmaupetit:
## Purpose

Current "plugins" implementation is far from academic. We should follow Ansible's recommendations to improve them.

## Proposal

- [x] Switch the app lookup call to a real lookup plugin
- [x] Fix merge filter implementation (remove `deepmerge` and make our own)
- [x] Add tests for plugins
- [x] Automate tests execution in the CI
- [x] Add python code linters (pylint, flake8, isort)


.circleci/config.yml
```diff
@@ -47,16 +55,17 @@ jobs:
             bin/build
             mkdir -p /tmp/docker/images
             docker save -o /tmp/docker/images/arnold.tar "${ARNOLD_IMAGE}"
-      # Store the built as an artifact that can be reused in other jobs
+      # Save built to the workspace so that it can be reused in other jobs
```
sampaccoud: s/built/build?

.circleci/config.yml
```diff
@@ -75,15 +85,29 @@ jobs:
...
+
+    steps:
+      - checkout
+      - *dind
+      - *ci_env
+      - *attach_workspace
+      - *docker_load
+      - run: bin/lint plugins
+
+  test-built:
```
sampaccoud, line 108: s/built/build?

.circleci/config.yml
```diff
@@ -105,22 +129,54 @@ jobs:
               "${ARNOLD_IMAGE}" \
               ansible --version
 
+  # FIXME: we have deactivated plugins test coverage as the container user is
+  # not allowed to write to /app/.coverage and CircleCI does not allowed to
+  # mount volumes with "docker run" commands.
```
sampaccoud: you're saying this and then below you're mounting /app/.pytest_cache?

.circleci/config.yml
```diff
@@ -105,22 +129,54 @@ jobs:
               "${ARNOLD_IMAGE}" \
               ansible --version
 
+  # FIXME: we have deactivated plugins test coverage as the container user is
+  # not allowed to write to /app/.coverage and CircleCI does not allowed to
```
sampaccoud: s/allowed/allow

.circleci/config.yml
```diff
@@ -105,22 +129,54 @@ jobs:
...
             - build
-      - docker_lint:
+      - lint-docker:
           requires:
             - build
-      - test:
+      - lint-plugins:
+          requires:
+            - build
+      - test-built:
```
sampaccoud, line 171: ok... maybe you really meant built instead of build...

bin/lint
```diff
@@ -47,10 +47,25 @@ _bash() {
...
+
+  echo -e "\n---> Plugins: isort\n"
+  _docker_run isort --recursive --diff --check-only plugins
+  print_green "  -> isort check OK"
+
+  echo -e "\n---> Plugins: Pylint\n"
+  _docker_run pylint plugins
+  print_green "  -> Pylint check OK"
+}
+
```
sampaccoud: no black?

plugins/filter/merge.py
```diff
@@ -5,41 +5,59 @@
...
-    for service in result["services"]:
-        new_app_selected_services = filter(
-            lambda s: s.get("name") == service.get("name"), new["services"]
-        )
+    for base_service in result["services"]:
+        new_app_selected_services = [
+            s
+            for s in new["services"]
+            if s.get("name") == base_service.get("name")
+        ]
```
sampaccoud: why did you switch from filter to list comprehension?

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def deep_sort_dict(d):
+    """
+        Recursively alphabetically sort lists as dictionary values (input dict
+        is not modified)
+    """
+    for k, v in d.iteritems():
+        if isinstance(v, list):
+            for i in v:
```
sampaccoud: when I see "i" I expect an integer

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+from ansible.compat.tests import unittest
+from ansible.errors import AnsibleFilterError
+
+from plugins.filter.merge import merge_with_app
+
+
+def deep_sort_dict(d):
+    """
+        Recursively alphabetically sort lists as dictionary values (input dict
+        is not modified)
```
sampaccoud: I think it is modified :thinking: 
You could start from an empty defaultdict and fill it recursively with what is returned by the next call?

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+        if isinstance(v, dict):
+            deep_sort_dict(v)
+    return d
+
+
+class TestDeepSortDict(unittest.TestCase):
+    """Test the ``deep_sort_dict`` utility"""
+
+    def test_flat_dict(self):
+        """Avoid testing recusive calls"""
```
sampaccoud: s/recusive/recursive
Actually I don't understand how this docstring describes what you are testing.

Maybe something like: "The deep_sort_dict method should sort flat (non nested) dictionaries as expected" ?

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+    def test_embedded_dict(self):
+        """Test recursive calls"""
+
+        d = {"foo": {"baz": [4, 5, 6, 1]}, "bar": ["c", "b", "d", "a"]}
+        expected = {"foo": {"baz": [1, 4, 5, 6]}, "bar": ["a", "b", "c", "d"]}
+
+        assert deep_sort_dict(d) == expected
+
+    def test_deep_embedded_dict(self):
+        """Avoid testing recursivity"""
```
sampaccoud: wrong docstring

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+        d = {
+            "foo": {"baz": [4, 1, 6, {"lol": ["x", "z", "y"]}]},
+            "bar": ["c", "b", "d", "a"],
+        }
+        expected = {
+            "foo": {"baz": [1, 4, 6, {"lol": ["x", "y", "z"]}]},
+            "bar": ["a", "b", "c", "d"],
+        }
+
+        assert deep_sort_dict(d) == expected
```
sampaccoud, line 76: you should add:

assert d == {
    "foo": {"baz": [4, 1, 6, {"lol": ["x", "z", "y"]}]},
    "bar": ["c", "b", "d", "a"],
}

Same on all tests?

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+class TestMergeWithAppFilter(unittest.TestCase):
+    """Tests for the ``merge_with_app`` filter"""
+
+    # View the full diff upon test failure
+    maxDiff = None
+
+    def assertDictDeepEqual(self, a, b):
+        """Use a sorted json string to deep compare apps (dictionaries)"""
```
sampaccoud: The docstring seems wrong. Here you are testing that they are pointing to the same reference, not that their json representation is the same.

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+            ({"name": ""}, {"name": "bar"}),
+        ):
+            with self.assertRaises(AnsibleFilterError) as cm:
+                merge_with_app(*args)
+            self.assertEqual(
+                cm.exception.message, "input apps should have the same name"
+            )
+
+    def test_submitted_apps_name_is_not_empty(self):
+        """Submitted apps should have the same ``name`` key"""
```
sampaccoud: copy/paste fail

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+
+        with self.assertRaises(AnsibleFilterError) as cm:
+            merge_with_app({"name": ""}, {"name": ""})
+        self.assertEqual(cm.exception.message, "input apps name cannot be empty")
+
+    def test_services_merge_with_empty_new_services(self):
+        """
+            Test app services merge with no services or volumes for tje new
```
sampaccoud: s/tje/the

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+            "volumes": ["volumes/foo_bar.yml", "volumes/foo_baz.yml"],
+        }
+
+        new = {"name": "foo", "services": []}
+
+        self.assertDictDeepEqual(merge_with_app(base, new), base)
+
+    def test_services_merge_without_duplicates(self):
+        """
+            Test app services merge does not creates duplicates if same items
```
sampaccoud: s/creates/create

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,258 @@
...
+                {
+                    "name": "bar",
+                    "configs": ["bar.conf", "bar2.conf"],
+                    "templates": ["bar/dc.yml", "bar/svc.yml", "bar/ep.yml"],
+                },
+                {
+                    "name": "baz",
+                    "configs": ["baz.conf"],
+                    "templates": ["baz/dc.yml", "baz/svc.yml", "baz/ep.yml"],
+                },
```
sampaccoud, line 239: So the fun service is gone is it what you expect?

.circleci/config.yml
jmaupetit: I meant save **the** built to the workspace.

.circleci/config.yml
jmaupetit: test the build :wink:

.circleci/config.yml
jmaupetit: It's not a classical volume it's a `tmpfs`. Concerning volumes, I think I cannot use them since we are in "dind" mode. We need to switch to the machine executor at least for this task.

.circleci/config.yml
jmaupetit: :thinking: it's the container user that is not **allowed** to

bin/lint
jmaupetit: `black` requires python 3.6+. I've tried this upgrade since Ansible 2.5+ is compatible with python 3 but we are running `debian:stretch` which is limited to python 3.5 :cry:  We can consider switching to an ubuntu-based image, but does it worth it?

.circleci/config.yml
sampaccoud: I nearly apologized... then I realized that there are 2 "allowed" in this sentence ^^

plugins/filter/merge.py
jmaupetit: Pylint told me to do so. I was annoyed when he told me that it was "the old way to do things" :flushed:

tests/units/plugins/filter/test_merge.py
jmaupetit: It was a rebase-related confusion. I've modified my implementation since to fit with this docstring :stuck_out_tongue:

tests/units/plugins/filter/test_merge.py
jmaupetit: For me `i` is an item, would you prefer `item`?

bin/lint
sampaccoud: If it's me I would switch to ubuntu because this container is just for tooling...

tests/units/plugins/filter/test_merge.py
jmaupetit: Right. I've fixed this docstring.

tests/units/plugins/filter/test_merge.py
jmaupetit: Yes. Sorry for that.

tests/units/plugins/filter/test_merge.py
jmaupetit: Not if we test immutability separately?

tests/units/plugins/filter/test_merge.py
jmaupetit: Wow. I should have carefully check my docstring. This is an old dumb implementation.

tests/units/plugins/filter/test_merge.py
jmaupetit: Yes, we take data from right (new) to left (base) if the app is defined in left (base), _i.e._ you cannot define a new app that is not listed as available (with files in apps/).

.circleci/config.yml
jmaupetit: Oops. you're right.

bin/lint
jmaupetit: Ok I'll declare a new issue for this.

tests/units/plugins/filter/test_merge.py
sampaccoud: For the moment I just read the tests. Maybe it deserves a line in the docstring. 
Then the concept could be explained in some documentation (in another PR is fine)?

.circleci/config.yml
sampaccoud: So why can't you use a tmpfs for /app/.coverage also?

jmaupetit: Because it's a file, not a directory.

bin/lint
jmaupetit: Done in #78 & #79

tests/units/plugins/filter/test_merge.py
```diff
@@ -0,0 +1,281 @@
...
+class TestDeepSortDict(unittest.TestCase):
+    """Test the ``deep_sort_dict`` utility"""
+
+    def test_immutability(self):
+        """Test that our implementation does not modify submitted dict in place"""
+
+        d = {"foo": 1, "bar": ["c", "b", "d", "a"]}
+        s = deep_sort_dict(d)
+
+        assert d != s
```
sampaccoud, line 45: assertion is good but you didn't test that it does not modify d :-)
assert d == {"foo": 1, "bar": ["c", "b", "d", "a"]} ?

plugins/lookup/apps.py
```diff
@@ -0,0 +1,184 @@
...
+        applications' tree, e.g. for the "richie" application and the
+        "apps" ``apps_path``:
+
+            apps/richie
+            ├── templates
+            │   ├── elasticsearch
+            │   │   ├── dc.yml.j2
+            │   │   └── svc.yml.j2
+            │   ├── nginx
+            │   │   ├── _configs
```
sampaccoud, line 33: I think singular is better: config

plugins/filter/merge.py
```diff
@@ -5,11 +5,13 @@
...
 
 from ansible.errors import AnsibleFilterError
-from deepmerge import always_merger
 
 
 def merge_with_app(base, new):
-    """Merge data from the "new" application to the "base" application"""
+    """
+        Merge data from the "new" application to the "base" application.
+        Services listed in "new" that does not exist in "base" will be ignored.
```
sampaccoud: s/does/do

plugins/filter/merge.py
```diff
@@ -21,31 +23,42 @@ def merge_with_app(base, new):
...
+        ]
+
         if len(new_app_selected_services) != 1:
             continue
-        service = always_merger.merge(service, new_app_selected_services[0])
+        new_service = new_app_selected_services[0]
+        for k in ("configs", "templates"):
+            if new_service.get(k) is None:
+                continue
+            base_service[k] = sorted(list(set(new_service[k] + base_service[k])))
```
sampaccoud: maybe add a comment to explain that you do list(set( to deduplicate?

sampaccoud (approved):

plugins/lookup/apps.py
jmaupetit: I see plural everywhere (volumes, templates, etc). Let's stick to it?

sampaccoud: but config is "the config"? Pick what you prefer.

jmaupetit: There may be more that once configuration for a service (_e.g._ edxapp cms + lms ngnix configurations).

PR #83, opened 2018-07-16
Add openshift to circleci

sampaccoud:
## Purpose

We want to be able to run our playbooks in the CI to really test deployments.

## Proposal

The easiest way to run OpenShift is to use the Docker image.
It requires a minor configuration of the Docker engine to allow an insecure registry.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,34 @@ jobs:
               "${ARNOLD_IMAGE}" \
               ansible --version
 
+  # Test the bootstrap playbook on a real OpenShift cluster installed in CircleCI's VM.
+  test-bootstrap:
+    machine: true
+    working_directory: ~/fun
+    steps:
+      - checkout
+      - *attach_workspace
+      - *docker_load
+      - *ci_env
+      - run:
+          name: Install the OC client
+          command: |
+            wget https://github.com/openshift/origin/releases/download/v3.9.0/openshift-origin-client-tools-v3.9.0-191fece-linux-64bit.tar.gz -P /tmp/openshift/
+            tar xzf /tmp/openshift/openshift*.tar.gz --strip-components=1 -C /tmp/openshift/
+            cp /tmp/openshift/oc $HOME/bin/
+      - run:
+          name: Configure Docker insecure registries for OpenShift cluster
+          command: |
+            sudo bash -c "echo '{\"insecure-registries\": [\"172.30.0.0/16\"]}' > /etc/docker/daemon.json"
+            sudo service docker restart
+      - run:
+          name: Bootstrap the test applications in OpenShift
+          command: |
+            LOCALE_IP=$(ip -o addr show eth0 | grep -v inet6 | sed "s/^.* inet \([0-9.]*\)\/[0-9]* .*$/\1/")
+            oc cluster up --server-loglevel=5 --public-hostname="${LOCALE_IP}"
+            oc login -u developer -p developer https://${LOCALE_IP}:8443 --insecure-skip-tls-verify=true
+            echo "arnold" | USE_TTY=false ./bin/bootstrap
+
   # FIXME: we have deactivated plugins test coverage as the container user is
   # not allowed to write to /app/.coverage and CircleCI does not allow to mount
   # volumes with "docker run" commands.
@@ -169,6 +197,12 @@ workflows:
           requires:
             - build
       - test-built:
+          requires:
+          - lint-bash
+          - lint-docker
+          - lint-ansible
+          - lint-plugins
+      - test-bootstrap:
           requires:
           - lint-bash
           - lint-docker
--- a/bin/_config.sh
+++ b/bin/_config.sh
@@ -35,8 +35,6 @@ function _set_minishift_path() {
 #     command
 function _docker_run() {
 
-    _set_minishift_path
-
     local env_file="env.d/development"
     local args
     local i=1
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -129,6 +129,27 @@ jobs:
...
+      - *attach_workspace
+      - *docker_load
+      - *ci_env
+      - run:
+          name: Test deploy in OpenShift
+          command: |
+            sudo bash -c "echo '{\"insecure-registries\": [\"172.30.0.0/16\"]}' > /etc/docker/daemon.json"
+            sudo service docker restart
+            wget https://github.com/openshift/origin/releases/download/v3.9.0/openshift-origin-client-tools-v3.9.0-191fece-linux-64bit.tar.gz
+            tar -xf openshift*.tar.gz --strip-components=1
```
jmaupetit: Why not using `tar xzf openshift*.tar.gz`? In a temporary directory and make an install (copy `oc`) in `$HOME/bin`?

.circleci/config.yml
```diff
@@ -129,6 +129,27 @@ jobs:
...
+    steps:
+      - checkout
+      - *attach_workspace
+      - *docker_load
+      - *ci_env
+      - run:
+          name: Test deploy in OpenShift
+          command: |
+            sudo bash -c "echo '{\"insecure-registries\": [\"172.30.0.0/16\"]}' > /etc/docker/daemon.json"
+            sudo service docker restart
```
jmaupetit: I would prefer making the configuration after the installation of `oc`. WDYT?

.circleci/config.yml
```diff
@@ -129,6 +129,27 @@ jobs:
...
+            wget https://github.com/openshift/origin/releases/download/v3.9.0/openshift-origin-client-tools-v3.9.0-191fece-linux-64bit.tar.gz
+            tar -xf openshift*.tar.gz --strip-components=1
+            export PATH="${PATH}:${PWD}"
+            LOCALE_IP=$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)
```
jmaupetit: Is the following easier to understand?

```bash
LOCALE_IP=$(ip -o addr show eth0 | grep -v inet6 | sed "s/^.* inet \([0-9.]*\)\/[0-9]* .*$/\1/")
```

.circleci/config.yml
```diff
@@ -129,6 +129,27 @@ jobs:
...
+            oc cluster up --server-loglevel=5 --public-hostname="${LOCALE_IP}"
+            oc login -u developer -p developer https://${LOCALE_IP}:8443 --insecure-skip-tls-verify=true
+            echo "arnold" | USE_TTY=false ./bin/bootstrap -e "apps_filter=richie"
```
jmaupetit: I think you can split this step in 3 separated steps:

1. Install & configure oc
2. Start the OpenShift cluster
3. Bootstrap the test application

bin/_config.sh
```diff
@@ -60,25 +58,24 @@ function _docker_run() {
...
-    if ${OC_LOGIN} && ! minishift status | grep Running > /dev/null ; then
-        echo "Error: minishift is not running. Please start it first."
-        exit 1
-    fi
-
     # Check that the current user is already logged in minishift
-    if ${OC_LOGIN} && ! oc whoami &> /dev/null; then
-        echo "Error: you need to login to minishift first."
+    if ! oc whoami &> /dev/null; then
+        echo "Error: you need to login to openshift first."
```
jmaupetit: OpenShift

bin/_config.sh
```diff
@@ -60,25 +58,24 @@ function _docker_run() {
...
         exit 1
     fi
 
-    docker run --rm -it \
+    OPENSHIFT_URL=$(oc version|grep "Server "|awk -F'Server ' '{print $2}')
```
jmaupetit: Can you add spaces between pipes to increase readability?

jmaupetit: Default awk separator is space, hence, this row can be simplified as follow:

```bash
OPENSHIFT_URL=$(oc version | grep Server | awk '{print $2}')
```

bin/_config.sh
```diff
@@ -60,25 +58,24 @@ function _docker_run() {
...
+
+    docker run --rm -i${TTY_OPTION} \
         -u "$(id -u)" \
         --env-file "$env_file" \
         --env OC_LOGIN \
         --env K8S_AUTH_API_KEY="$(oc whoami -t)" \
-        --env K8S_AUTH_HOST="https://$(minishift ip):8443" \
-        --env MINISHIFT_IP="$(minishift ip)" \
+        --env K8S_AUTH_HOST="${OPENSHIFT_URL}" \
+        --env OPENSHIFT_DOMAIN="${OPENSHIFT_DOMAIN}" \
```
jmaupetit: If it's the same variable name, you can pass it to docker run with: `--env OPENSHIFT_DOMAIN`

jmaupetit (commented):
That's an awesome feature. Thanks for having taken the time to implement it. :pray:

bin/_config.sh
```diff
@@ -60,25 +58,28 @@ function _docker_run() {
         i=$(( i+1 ))
     done
 
-    # Check that minishift is running
-    if ${OC_LOGIN} && ! minishift status | grep Running > /dev/null ; then
-        echo "Error: minishift is not running. Please start it first."
+    # Check that the current user is already logged in OpenShift
+    if ! oc whoami &> /dev/null; then
```
jmaupetit: Some scripts using Arnold (such as `pytest`) do not require to be logged in to OpenShift, and, thus are not impacted by the `K8S_AUTH_HOST` or `OPENSHIFT_DOMAIN` values. Hence, I think we must keep the `OC_LOGIN` flag functional:

```bash
if ${OC_LOGIN} && ! oc whoami &> /dev/null; then
    # ...
```

bin/_config.sh
```diff
@@ -60,25 +58,28 @@ function _docker_run() {
...
+        echo "Error: you need to login to an OpenShift server first."
         exit 1
     fi
 
-    # Check that the current user is already logged in minishift
-    if ${OC_LOGIN} && ! oc whoami &> /dev/null; then
-        echo "Error: you need to login to minishift first."
-        exit 1
-    fi
+    [[ "$USE_TTY" == "false" ]] && TTY_OPTION="" || TTY_OPTION="t"
```
jmaupetit: I would prefer: `TTY_OPTION="-t"` and adding a space between `-i` and the `${TTY_OPTION}` in the `docker run` invocation.

jmaupetit (approved):
I only have two minor changes (the latest is optional). After that: ship it!

bin/_config.sh
sampaccoud: ok

PR #84, opened 2018-07-17
Add default vars to apps and make them overridable

jmaupetit:
## Purpose

Each app should define it's own default vars to work, and those defaults should be overridable for a customer and an environment.

## Proposal

- [x] Scan the `vars/all` directory of each app to lookup defaults
- [x] Load default vars for each active app
- [x] Make defaults overridable for a customer/env combination (in `group_vars/customer/{{ customer }}/{{ env_type }}/*.yml`


tasks/set_vars.yml
```diff
@@ -1,11 +1,11 @@
 ---
 # tasks/set_vars.yml
 
-- name: Include customer env_type vars
+- name: Include global customer & env_type vars
   include_vars:
     file: "{{ item }}"
   with_items:
-    - "group_vars/customer/{{ customer }}.yml"
+    - "group_vars/customer/{{ customer }}/main.yml"
```
cchaudier, line 8: Why not all files in customer dir ?

cchaudier (approved):
Juste on question but you can merge if you wan.

tasks/set_vars.yml
jmaupetit: I think we need to control the importation sequence, WDYT?

PR #85, opened 2018-07-17
🎨(scripts) improve OpenShift's env setting

jmaupetit:
## Purpose

The `OC_LOGIN` flag to inactivate OpenShift's login via the `oc` client is too complex/ugly.

## Proposal

Define an `_set_openshift_env` routine and use it whenever we require to login to our OpenShift server.

### DIFF
--- a/bin/_config.sh
+++ b/bin/_config.sh
@@ -10,6 +10,34 @@ function _set_minishift_path() {
     export PATH="$MINISHIFT_PATH:$PATH"
 }
 
+# _set_openshift_env: set OpenShift's environment
+#
+# Use this function to configure the OpenShift environment whenever your docker
+# run commands require access to an OpenShift server. Nota bene: you'll need to
+# login to your OpenShift server first.
+function _set_openshift_env() {
+
+    # Check that the current user is already logged in OpenShift
+    if ! oc whoami &> /dev/null; then
+        echo "Error: you need to login to an OpenShift server first."
+        exit 1
+    fi
+
+    # Ansible's OpenShift raw module requires the following two environment
+    # variables to be defined
+    K8S_AUTH_API_KEY="$(oc whoami -t)"
+    export K8S_AUTH_API_KEY
+    K8S_AUTH_HOST=$(oc version | grep Server | awk '{print $2}')
+    export K8S_AUTH_HOST
+
+    # The OpenShift/k8s host is of the form https://domain:8443 so we can
+    # extract the domain from it
+    #
+    # shellcheck disable=SC2034,SC2001
+    OPENSHIFT_DOMAIN=$(echo "${K8S_AUTH_HOST}" | sed 's#https://\(.*\):8443#\1#')
+    export OPENSHIFT_DOMAIN
+}
+
 # _docker_run: wrap docker run command
 #
 # usage: _docker_run [OPTIONS] [COMMAND]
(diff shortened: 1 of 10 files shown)


bin/_config.sh
```diff
@@ -58,28 +79,14 @@ function _docker_run() {
...
-
     docker run --rm -i ${TTY_OPTION} \
         -u "$(id -u)" \
         --env-file "$env_file" \
-        --env OC_LOGIN \
-        --env K8S_AUTH_API_KEY="$(oc whoami -t)" \
-        --env K8S_AUTH_HOST="${OPENSHIFT_HOST}" \
+        --env K8S_AUTH_API_KEY \
+        --env K8S_AUTH_HOST \
+        --env OPENSHIFT_HOST \
```
sampaccoud: redundant with K8S_AUTH_HOST?

sampaccoud (commented):
Just a comment on rationalizing 2 variables that seem redundant

bin/_config.sh
jmaupetit: Indeed, but I find it more explicit, and the other one is required. Should I remove it?

bin/_config.sh
```diff
@@ -10,6 +10,36 @@ function _set_minishift_path() {
...
+    # The OpenShift host is of the form https://domain:8443 so we can
+    # extract the domain from it
+    # shellcheck disable=SC2034,SC2001
+    OPENSHIFT_DOMAIN=$(echo "${OPENSHIFT_HOST}" | sed 's#https://\(.*\):8443#\1#')
+    export OPENSHIFT_DOMAIN
+
+    # Ansible's OpenShift raw module requires the following two environment
+    # variables to be defined
+    K8S_AUTH_API_KEY="$(oc whoami -t)"
+    export K8S_AUTH_API_KEY
```
cchaudier, line 29: You can concat this 2 lines : 
`export K8S_AUTH_API_KEY="$(oc whoami -t)"`

bin/_config.sh
```diff
@@ -10,6 +10,36 @@ function _set_minishift_path() {
...
+
+    # Check that the current user is already logged in OpenShift
+    if ! oc whoami &> /dev/null; then
+        echo "Error: you need to login to an OpenShift server first."
+        exit 1
+    fi
+
+    # Get the OpenShift host from OC
+    OPENSHIFT_HOST=$(oc version | grep Server | awk '{print $2}')
+    export OPENSHIFT_HOST
```
cchaudier: Why don't use K8S_AUTH_HOST instead ?

bin/_config.sh
jmaupetit: I've removed it.

bin/lint
```diff
@@ -11,7 +11,6 @@ _docker_run() {
     local docker_volume_opts=""
   fi
   docker run --rm -it \
-    -e OC_LOGIN=false \
```
cchaudier, line 14: I think you must also change the `.circleci/config.yml` with this change.

cchaudier (commented):
Great.
I'v some comments.

bin/_config.sh
jmaupetit: That was my first implementation, but `shellcheck` advises me not to declare and initialize in the same statement.

bin/lint
jmaupetit: Done.

bin/_config.sh
jmaupetit: Done.

bin/_config.sh
cchaudier: Oh no :'(

cchaudier (approved):

cchaudier:
Ship it

PR #88, opened 2018-07-19
Move Richie's application to the new application base logic

jmaupetit:
## Purpose

Now that we've prepared everything to make deployable applications pluggable, it's time to move distribute default application vars with the application itself and polish/fix related templates.

## Proposal

Concerning Richie application:

- [x] fix app/services labels in templates
- [x] define an host+port for the patient0/development project
- [x] move Richie's variables from `group_vars/all/main.yml` to `apps/richie/vars/all/main.yml` and use them in templates
- [x] test Richie's deployment in the CI


.circleci/config.yml
```diff
@@ -155,30 +155,53 @@ jobs:
...
+            echo "export OPENSHIFT_DOMAIN=${OPENSHIFT_DOMAIN}" >> $BASH_ENV
+            echo "export K8S_AUTH_HOST=${K8S_AUTH_HOST}" >> $BASH_ENV
+            echo "export K8S_AUTH_API_KEY=${K8S_AUTH_API_KEY}" >> $BASH_ENV
+            source $BASH_ENV
 
+      - run:
+          name: Create applications bootstraping routine
+          command: |
+            cat << EOF > bin/ci-bootstrap
+              #!/usr/bin/env/bash
```
sampaccoud: Then this should be in bin/ as a script...

sampaccoud: I think we need much more comments on all this.

apps/richie/templates/postgresql/svc.yml.j2
```diff
@@ -2,25 +2,28 @@ apiVersion: v1
...
       protocol: TCP
-      targetPort: {{ postgresql.port }}
+      targetPort: {{ richie_postgresql_port }}
 # As commented in the ad hoc endpoint, the endpoint name points to this service
 # so that it does not rely on a deployment configuration when the "env_type" is
 # not "development" or "feature" (not yet implemented). In environments other
 # than "development" or "feature", we use a PostgreSQL cluster outside of
 # OpenShift.
-{% if env_type == "development"  %}
+{% if env_type in ("ci", "development", "feature")  %}
```
sampaccoud: @cchaudier @rouja is there a way to define constants so that "development", "ci", "feature" are defined only once for all?

apps/richie/vars/all/main.yml
```diff
@@ -0,0 +1,28 @@
...
+richie_host: "richie.{{ project_name}}.{{ domain_name }}"
+
+# -- nginx
+richie_nginx_image_name: "fundocker/openshift-nginx"
+richie_nginx_image_tag: "1.13"
+richie_nginx_port: 8061
+
+# -- elasticsearch
+richie_elasticsearch_image_name: "fundocker/openshift-elasticsearch"
+richie_elasticsearch_image_tag: "6.2.4"
```
sampaccoud, line 13: we moved to 6.3

create_project.yml
```diff
@@ -17,3 +20,7 @@
         description: "{{ project_description|default(project_display_name)|default(project_name) }}"
         display_name: "{{ project_display_name|default(project_name) }}"
         state: present
+      register: new_project
+      until: new_project.result is defined and new_project.result.status is success
+      retries: 30
+      delay: 10
```
sampaccoud, line 26: so that's all we need to do to make it synchronous?

group_vars/customer/patient0/ci/secrets/edxapp.vault.yml
```diff
@@ -0,0 +1,42 @@
+$ANSIBLE_VAULT;1.1;AES256
```
sampaccoud: this is related to edxapp. Fail?

templates/openshift/common/utils/get_deployment_stamp_from_route.yml.j2
```diff
@@ -1,7 +1,9 @@
 apiVersion: v1
 kind: Route
 metadata:
-  name: "{{ app.name }}-{{ prefix }}"
+  # We only keep track of the first hostname found as we want to fetch the
+  # deployment_stamp that is common to every routes for a deployment
```
sampaccoud: I don't understand this comment.
What are the other hostnames that I could get?

.circleci/config.yml
jmaupetit: I wanted to maintain a self-contained CI script. I'll add comments to explain what it does. In future optimizations, maybe we can extract this in a sugar script to make things simpler.

apps/richie/templates/postgresql/svc.yml.j2
sampaccoud: maybe this? https://devops.stackexchange.com/questions/3474/how-to-define-constants-in-ansible

apps/richie/vars/all/main.yml
jmaupetit: Is there compatibility issues with the Richie's image we are using? I think it can be treated in a new PR upgrading Richie to its latest release. WDYT?

create_project.yml
jmaupetit: Kind of. This ensures that we don't get a 409 response while trying to create a project that is not fully deleted.

group_vars/customer/patient0/ci/secrets/edxapp.vault.yml
jmaupetit: :thinking: Yup. I shouldn't have add this secret on this PR.

create_project.yml
sampaccoud: then we could use the same pattern to make jobs wait for edxapp to be up?

apps/richie/vars/all/main.yml
sampaccoud: I didn't run into any issue when upgrading. Yes this can be postponed. It raises the question of how to best keep this in sync with what we do in the richie repository... :thinking:

templates/openshift/common/utils/get_deployment_stamp_from_route.yml.j2
jmaupetit: An application can define several routes (_e.g._ two services with two different routes). Our goal here is to get the `deployment_stamp` of a deployed stack by getting a route object. As the `deployment_stamp` is commonly shared among objects of a stack, there is no need to get more than one route to achieve this. Is that more clear?

sampaccoud: Ok! The key to understanding is "Our goal here is to get the deployment_stamp of a deployed stack by getting a route object". Can you insert your whole answer as a comment?

group_vars/customer/patient0/ci/secrets/edxapp.vault.yml
jmaupetit: I've removed it. Sorry for that.

create_project.yml
jmaupetit: You read my mind, it's already planned :wink: I'll add dedicated tasks to run jobs. But still, we will need to improve jobs ordering/dependency.

apps/richie/vars/all/main.yml
jmaupetit: As long as we maintain deployable apps in Arnold's repository, we'll need to make a new release for Arnold each time we upgrade one of those apps.

apps/richie/templates/postgresql/svc.yml.j2
jmaupetit: What about the following?

```yaml
# group_vars/all/main.yml
trashable_env_types:
  - "ci"
  - "development"
  - "feature"

# apps/richie/templates/postgresql/svc.yml.j2
{% if env_type in trashable_env_types %}
# ...
```

apps/richie/vars/all/main.yml
sampaccoud: I was thinking of a unique place in richie's repository where the ES image is referenced and we get it from this place for:
- docker-compose
- arnold
- etc.

apps/richie/templates/postgresql/svc.yml.j2
sampaccoud: And also the answer to this https://devops.stackexchange.com/questions/3474/how-to-define-constants-in-ansible advising to use `defaults/main.yml` to define some kind of "constants". The idea is to use this constant everywhere we wrote "development". WDYT?

apps/richie/vars/all/main.yml
jmaupetit: This place can be the docker-compose file and Arnold _et al._ can stick to it. Do you have something more automated in mind?

apps/richie/templates/postgresql/svc.yml.j2
jmaupetit: `group_vars/all/main.yml` are our defaults that we can override.

> The idea is to use this constant everywhere we wrote "development".

Ok to use it for service templates (these are the only place where we use this condition).

apps/richie/vars/all/main.yml
sampaccoud: ok. I didn't think it would be possible/easy to retrieve the info from a docker-compose file.

sampaccoud (approved):

apps/richie/vars/all/main.yml
jmaupetit: :thinking: I am not sure to understand what you have in mind, but we'll discuss this IRL before declaring a new issue.

PR #93, opened 2018-07-30
🐛(objects) make objects name unique

jmaupetit:
## Purpose

When two (or more) objects from different applications share the same name, objects conflict with each others event if their labels differ.

## Proposal

- [x] Prefix route by their related application name to avoid cross-apps collisions in route names
- [x] Prefix all applications objects by their application name

### DIFF
--- a/apps/richie/templates/richie/dc.yml.j2
+++ b/apps/richie/templates/richie/dc.yml.j2
@@ -6,20 +6,20 @@ metadata:
     service: richie
     version: "{{ richie_image_tag }}"
     deployment_stamp: "{{ deployment_stamp }}"
-  name: richie-{{ deployment_stamp }}
+  name: "richie-richie-{{ deployment_stamp }}"
   namespace: "{{ project_name }}"
 spec:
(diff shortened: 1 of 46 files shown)


sampaccoud (approved):

jmaupetit:
It seems that the collision occurs for every OS objects, not only routes. I'll update this PR description.

apps/richie/templates/richie/dc.yml.j2
```diff
@@ -6,20 +6,20 @@ metadata:
     service: richie
     version: "{{ richie_image_tag }}"
     deployment_stamp: "{{ deployment_stamp }}"
-  name: richie-{{ deployment_stamp }}
+  name: "richie-richie-{{ deployment_stamp }}"
```
sampaccoud: Arf... Remember when I told you that the service should be called "app" and not richie. I think you really need to consider that. And it's more natural to follow the docker-compose semantic...

jmaupetit: Ok, you are right: let's do this.

PR #95, opened 2018-07-30
Add celery workers to LMS and CMS for edxapp

sampaccoud:
## Purpose

We need to run Celery workers to handle asynchronous tasks.

## Proposal

- [x] add Redis as a separate app (to avoid redeploying it each time we redeploy edxapp) and without blue/green (because Redis does not support several instances on the same volume...) 
- [x] add Celery workers to edxapp (one for each queue in both LMS and CMS)
- [x] remove the "selector" item in dc templates as it seems useless and is generating errors with multiple deployments per service.
- [x] add vars for celery queues and use them in DCs + settings

### DIFF
--- a/apps/edxapp/templates/cms/_configs/settings.yml.j2
+++ b/apps/edxapp/templates/cms/_configs/settings.yml.j2
@@ -8,11 +8,10 @@ LMS_BASE: "{{ edxapp_lms_host }}"
 CMS_BASE: "{{ edxapp_cms_host }}"
 
 # Celery Broker
-# FIXME: we should use the redis_app_* default vars
 CELERY_BROKER_TRANSPORT: redis
-CELERY_BROKER_HOST: redis
+CELERY_BROKER_HOST: "{{ redis_app_host }}"
 CELERY_BROKER_VHOST: 0
-CELERY_BROKER_PORT: 6379
+CELERY_BROKER_PORT: "{{ redis_app_port }}"
 # FIXME: for now, we haven't set a user/password to connect to redis. Once those
 # are set, you'll need to move the two following settings in edxapp secret
 CELERY_BROKER_USER: ""
--- a/apps/edxapp/templates/cms/_configs/settings.yml.j2
+++ b/apps/edxapp/templates/cms/_configs/settings.yml.j2
@@ -16,3 +16,8 @@ CELERY_BROKER_PORT: "{{ redis_app_port }}"
 # are set, you'll need to move the two following settings in edxapp secret
 CELERY_BROKER_USER: ""
 CELERY_BROKER_PASSWORD: ""
+
+# Celery queues
+HIGH_PRIORITY_QUEUE: "{{ edxapp_celery_cms_high_priority_queue }}"
+DEFAULT_PRIORITY_QUEUE: "{{ edxapp_celery_cms_default_priority_queue }}"
+LOW_PRIORITY_QUEUE: "{{ edxapp_celery_cms_low_priority_queue }}"
--- a/apps/edxapp/templates/cms/dc.yml.j2
+++ b/apps/edxapp/templates/cms/dc.yml.j2
@@ -10,9 +10,6 @@ metadata:
   namespace: "{{ project_name }}"
 spec:
   replicas: 1  # number of pods we want
-  selector:
-    app: edxapp
-    deploymentconfig: edxapp-cms-{{ deployment_stamp }}
   template:
     metadata:
       labels:
--- a/apps/edxapp/templates/lms/_configs/settings.yml.j2
+++ b/apps/edxapp/templates/lms/_configs/settings.yml.j2
@@ -16,3 +16,9 @@ CELERY_BROKER_PORT: "{{ redis_app_port }}"
 # are set, you'll need to move the two following settings in edxapp secret
 CELERY_BROKER_USER: ""
 CELERY_BROKER_PASSWORD: ""
+
+# Celery queues
+HIGH_PRIORITY_QUEUE: "{{ edxapp_celery_lms_high_priority_queue }}"
+DEFAULT_PRIORITY_QUEUE: "{{ edxapp_celery_lms_default_priority_queue }}"
+LOW_PRIORITY_QUEUE: "{{ edxapp_celery_lms_low_priority_queue }}"
+HIGH_MEM_QUEUE: "{{ edxapp_celery_lms_high_mem_queue }}"
--- a/apps/edxapp/templates/lms/_configs/settings.yml.j2
+++ b/apps/edxapp/templates/lms/_configs/settings.yml.j2
@@ -8,11 +8,10 @@ LMS_BASE: "{{ edxapp_lms_host }}"
 CMS_BASE: "{{ edxapp_cms_host }}"
 
 # Celery Broker
-# FIXME: we should use the redis_app_* default vars
 CELERY_BROKER_TRANSPORT: redis
-CELERY_BROKER_HOST: redis
+CELERY_BROKER_HOST: "{{ redis_app_host }}"
 CELERY_BROKER_VHOST: 0
-CELERY_BROKER_PORT: 6379
+CELERY_BROKER_PORT: "{{ redis_app_port }}"
 # FIXME: for now, we haven't set a user/password to connect to redis. Once those
 # are set, you'll need to move the two following settings in edxapp secret
 CELERY_BROKER_USER: ""
--- a/apps/edxapp/templates/lms/dc_queue_high_priority.yml.j2
+++ b/apps/edxapp/templates/lms/dc_queue_high_priority.yml.j2
@@ -1,5 +1,5 @@
 {% set service_variant = "lms" %}
-{% set worker_type = "wsgi" %}
-{% set queue = "" %}
+{% set worker_type = "queue" %}
+{% set queue = edxapp_celery_lms_high_priority_queue %}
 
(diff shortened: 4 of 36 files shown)


jmaupetit:
In your commit message: 

> **sparkles(edxapp) add redis app**
>
> [...] Note that blue/green is note activated on Redis because only one container at a time should be mounted on the Redis data volume.

`s/note/not/`

apps/redis/templates/app/svc.yml.j2
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: redis
+    service: app
+    version: "{{ redis_app_image_tag }}"
+  name: "{{ redis_app_host }}"  # name of the service should be host name in nginx
```
jmaupetit: Can you explain this comment about `nginx`? Do you mean in `edxapp` settings?

apps/edxapp/templates/cms/_configs/settings.yml.j2
```diff
@@ -6,3 +6,8 @@ SITE_NAME: "{{ project_display_name }}"
 
 LMS_BASE: "{{ edxapp_lms_host }}"
 CMS_BASE: "{{ edxapp_cms_host }}"
+
+# Celery Broker
+CELERY_BROKER_TRANSPORT: redis
+CELERY_BROKER_HOSTNAME: redis
```
jmaupetit: Should be `"{{ redis_app_host }}"`?

apps/edxapp/templates/cms/dc_worker_default_priority.yml.j2
```diff
@@ -0,0 +1,7 @@
+{% extends "dc.yml.j2" %}
+{% block command %}
+        command:
+          - "/bin/bash"
+          - "-c"
+          - python manage.py lms celery worker --loglevel=info --queues=edx.lms.core.high --hostname=edx.lms.core.default.%%h --concurrency=1
+{% endblock %}
```
jmaupetit: A carriage return is missing at the end of all dc_worker* templates.

apps/edxapp/templates/cms/dc_worker_default_priority.yml.j2
```diff
@@ -0,0 +1,7 @@
+{% extends "dc.yml.j2" %}
+{% block command %}
+        command:
+          - "/bin/bash"
+          - "-c"
+          - python manage.py lms celery worker --loglevel=info --queues=edx.lms.core.high --hostname=edx.lms.core.default.%%h --concurrency=1
```
jmaupetit: I am not familiar with this command, but: it seems that the hostname matches the default host, but you are always using the `edx.lms.core.high` queue (the same remark applies to other worker templates). Is that expected?

apps/edxapp/templates/cms/dc_worker_low_priority.yml.j2
```diff
@@ -0,0 +1,7 @@
+{% extends "dc.yml.j2" %}
+{% block command %}
+        command:
+          - "/bin/bash"
+          - "-c"
+          - python manage.py lms celery worker --loglevel=info --queues=edx.lms.core.high --hostname=edx.lms.core.low.%%h --concurrency=1
```
jmaupetit: Those 3 templates should target the CMS not the LMS: `python manage.py cms [...] --queues=edx.cms.core.xx --hostname=edx.cms.core.low.%%h [...]`

apps/edxapp/templates/lms/_configs/settings.yml.j2
```diff
@@ -6,3 +6,8 @@ SITE_NAME: "{{ project_display_name }}"
 
 LMS_BASE: "{{ edxapp_lms_host }}"
 CMS_BASE: "{{ edxapp_cms_host }}"
+
+# Celery Broker
+CELERY_BROKER_TRANSPORT: redis
+CELERY_BROKER_HOSTNAME: redis
```
jmaupetit: Should be `"{{ redis_app_host }}"`?

apps/edxapp/templates/lms/dc_queue_high_priority.yml.j2
```diff
@@ -0,0 +1,7 @@
+{% extends "dc.yml.j2" %}
+{% block command %}
+        command:
+          - "/bin/bash"
+          - "-c"
+          - python manage.py lms celery worker --loglevel=info --queues=edx.lms.core.high --hostname=edx.lms.core.high.%%h --concurrency=1
```
jmaupetit: Same remark about targeted queues.

apps/edxapp/templates/cms/dc.yml.j2
```diff
@@ -10,9 +10,6 @@ metadata:
   namespace: "{{ project_name }}"
 spec:
   replicas: 1  # number of pods we want
-  selector:
-    app: edxapp
-    deploymentconfig: edxapp-cms-{{ deployment_stamp }}
```
jmaupetit: :+1:

apps/edxapp/templates/cms/dc_worker_default_priority.yml.j2
sampaccoud: As discussed this morning, if you want to make this explicit, the right way to do it would be to declare the following settings explicitly in Arnold from Ansible variables: https://github.com/openfun/fun-platform/blob/master/config/lms/docker_run_production.py#L64

apps/edxapp/vars/all/main.yml
```diff
@@ -34,3 +34,12 @@ edxapp_nginx_image_name: "fundocker/openshift-nginx"
 edxapp_nginx_image_tag: "1.13"
 edxapp_nginx_cms_port: 8081
 edxapp_nginx_lms_port: 8071
+
+# -- celery/redis
+edxapp_celery_lms_high_priority_queue: "edx.lms.core.high"
+edxapp_celery_lms_default_priority_queue: "edx.lms.core.default"
+edxapp_celery_lms_low_priority_queue: "edx.lms.core.low"
+edxapp_celery_lms_high_mem_queue: "edx.lms.core.high_mem"
+edxapp_celery_cms_high_priority_queue: "edx.cms.core.high"
```
sampaccoud: maybe split in two groups lms and cms?

apps/edxapp/vars/all/main.yml
```diff
@@ -34,3 +34,12 @@ edxapp_nginx_image_name: "fundocker/openshift-nginx"
...
+edxapp_celery_cms_default_priority_queue: "edx.cms.core.default"
+edxapp_celery_cms_low_priority_queue: "edx.cms.core.low"
```
sampaccoud: In fact I realized that edX is using the same queues for lms and cms. should we still split?

tasks/manage_app.yml
```diff
@@ -14,7 +14,7 @@
 
 - name: Set OpenShift objects to manage
   set_fact:
-    deployments: "{{ templates | map('regex_search', '.*/dc*.yml.j2$') | select('string') | list }}"
+    deployments: "{{ templates | map('regex_search', '.*/dc.*.yml.j2$') | select('string') | list }}"
```
sampaccoud: will it take dc.yml.j2?

tasks/manage_app.yml
```diff
@@ -14,7 +14,7 @@
 
 - name: Set OpenShift objects to manage
   set_fact:
-    deployments: "{{ templates | map('regex_search', '.*/dc.yml.j2$') | select('string') | list }}"
+    deployments: "{{ templates | map('regex_search', '.*/dc.*.yml.j2$') | select('string') | list }}"
```
sampaccoud: will this find dc.yml.j2?

sampaccoud: Oh ok I just realized that this is a regex ! So you need to escape the other . no?
```
'.*/dc.*\.yml\.j2$'
```

apps/edxapp/vars/all/main.yml
jmaupetit: :thinking: I think we can also use the same queues. In this case, what about the workers? Should we keep two types of them?

apps/edxapp/vars/all/main.yml
jmaupetit: What do you mean by "group"?

tasks/manage_app.yml
jmaupetit: Dot (`.`) works fine :wink:, but true!

tasks/manage_app.yml
jmaupetit: Yes `.*` means any occurrence (of any character) even none.

apps/edxapp/vars/all/main.yml
sampaccoud: add a comment line to split in two groups... just visual

apps/edxapp/vars/all/main.yml
sampaccoud: so in fact edx IS using different queues. It is overriden in the aws.py setting file and I was looking at the common.py setting file. This being said, we could decide to run one worker per priority level and run both the lms and the cms queues on this worker... the setting when launching a worker is "queues" plural and can take several...

I'm for running 7 containers as it is simpler...

sampaccoud:
@jmaupetit Let's ship it! That's a great piece of team work. Thank you @rouja for kicking our ass on this one :rofl:

jmaupetit (approved):

PR #97, opened 2018-08-02
Improve routes management

### LINKED ISSUE #90: Add support for multiple host per-service
## Purpose

We currently only support one host per service definition to create underlying routes. As some apps (like `edxapp`) may have many routes pointing to the same service (_e.g._ nginx), we need to improve hosts management.

## Proposal

Move the `app.service.host` type from a simple string to a list of dictionaries with the route, the targeted service name and the targeted port.

jmaupetit:
## Purpose

Listing routes from the `apps` global variable definition was not that explicit nor easy to maintain. Hence, following our global "app-driven" logic, we decided to manage routes from their definition stated in an application service template. 

## Proposal

- [x] upgrade routes discovery and creation
- [x] remove hosts/ports definitions in the `apps` global var
- [x] add routes for every application service that requires one or more

As a consequence, this PR will fix #90.


apps/richie/templates/nginx/route.yml.j2
```diff
@@ -0,0 +1,43 @@
+{%- set host = richie_host -%}
+
+{#
+    Host prefix can be: previous, next or current. You can use this template with
+    either a prefix set using the following variables (in precedence order):
+
+        - item.1 (in a nested loop)
```
sampaccoud: I don't understand "in a nested loop"...

sampaccoud: can item be a namedtuple of a dictionary so we write something else than "item.1" !? Can item have a better name also ? it's an item of what?

jmaupetit: I'm referring to Ansible nested loops: https://docs.ansible.com/ansible/2.5/user_guide/playbooks_loops.html#complex-loops

But I got your point and I'll come with a better idea / implementation.

group_vars/customer/hello/main.yml
```diff
@@ -6,14 +6,6 @@ project_description: "A minimal project with a hello-world app ({{ env_type }})"
...
-    services:
-      - name:       hello
-        host:       "hello.{{ project_name }}.{{ domain_name }}"
-
-## _____ soon deprecated below _________________________________________________
-hello_host: "{{ project_name }}.{{ domain_name }}"
-hello_port: "8080"
-hello_msg:  "Hello OpenShift! by Arnold"
 
 nginx_ports:
```
cchaudier, line 10: This also  soon deprecated

group_vars/all/main.yml
```diff
@@ -34,6 +34,12 @@ endpoint_mongodb_ip:     192.168.10.11
 endpoint_mysql_ip:       192.168.10.254
 endpoint_postgresql_ip:  192.168.0.1
 
+# Blue/Green deployment route prefixes
+blue_green_route_prefixes:
+  - "previous"
+  - "current"
+  - "next"
```
cchaudier, line 41: I'v do this before but 
@sampaccoud on 12 Jun  ask this :
I'm skeptical that this should be configurable. Keep it simple and impose it?

See : https://github.com/openfun/arnold/pull/60

plugins/filter/deploy.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# pylint: disable=no-self-use,too-few-public-methods
+class FilterModule(object):
+    """Filters used for deployments"""
+
+    def filters(self):
+        """List plugin filters"""
+
+        return {"blue_green_host": blue_green_host}
```
cchaudier, line 46: This  filter is just for not using the nested loop ?

apps/edxapp/templates/nginx/route_cms.yml.j2
```diff
@@ -0,0 +1,25 @@
...
+    app: "edxapp"
+    service: "nginx"
+    version: "{{ edxapp_nginx_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+    route_prefix: "{{ prefix }}"
+    route_target_service: "cms"
+  annotations:
+    kubernetes.io/tls-acme: "true"
+spec:
+  host: "{{ edxapp_cms_host | blue_green_host(prefix) }}"
```
cchaudier, line 18: This this the host can't be overrided in the apps dict.
If whe want otherride this host we must declare `edxapp_cms_host` in the group_vars.

sampaccoud (approved):

cchaudier (commented):
Huge job ! :+1: 
I've somme comments

apps/edxapp/templates/nginx/route_cms.yml.j2
jmaupetit: >  If whe want otherride this host we must declare edxapp_cms_host in the group_vars.

Yes.

plugins/filter/deploy.py
jmaupetit: No this is to avoid adding logic to the template.

group_vars/all/main.yml
jmaupetit: :thinking: it might evolve...

cchaudier: Probably for me it's great like this ^^

plugins/filter/deploy.py
cchaudier: Ok in this case

apps/edxapp/templates/nginx/route_cms.yml.j2
cchaudier: Ok but I think the apps dict lose is first interest after your PR.

jmaupetit: I disagree and you will see why in the next PR.

Issue #34, opened 2018-05-28
Plug Datadog APM


cchaudier:
Add possibility to plug datadog APM to the apps.

PR #101, opened 2018-08-03
WIP: Plug datadog apm

cchaudier:
## Purpose
Add possibility to plug datadog APM to the apps. See #34.

## Proposal

We must override the official docker images.
To do this we create a BuildConfig to add datadog APM (ddtrace) on service images.
The DeployementConfig use the OpenShift builded image.
We do this only if th ddtrace ansible vars is set to true.

- [x] add richie templates
- [ ] add cms templates
- [ ] add lms templates

**edit**: this PR is depends on #111


apps/richie/templates/app/dc.yml.j2
```diff
@@ -38,6 +43,18 @@ spec:
...
+            - name: DATADOG_TRACE_AGENT_HOSTNAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_HOST
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_PORT
+              value: '8126'
```
jmaupetit, line 56: Should it be configurable, _e.g._ stored in a var?

apps/richie/templates/app/dc.yml.j2
```diff
@@ -20,8 +20,13 @@ spec:
         deployment_stamp: "{{ deployment_stamp }}"
     spec:
       containers:
-        - name: richie
+        - name: "richie"
+{% if ddtrace is defined and ddtrace %}
+          image: "docker-registry.default.svc:5000/{{ project_name }}/ddtrace-richie-{{ deployment_stamp }}:{{ richie_image_tag }}"
+          #image: "ddtrace-richie-{{ deployment_stamp }}:{{ richie_image_tag }}"
```
jmaupetit, line 26: Why do you need to force docker registry address? Because of `minishift`?

Anyway, only the commented line should stay right?

apps/richie/templates/app/bc.yml.j2
```diff
@@ -0,0 +1,35 @@
...
+  - type: ConfigChange
+  strategy:
+    type: Docker
+  source:
+    dockerfile: |-
+      FROM {{ richie_image_name }}:{{ richie_image_tag }}
+      # Switch back to the root user to install development dependencies
+      USER root:root
+
+      RUN pip install --no-cache-dir --prefix=/usr/local ddtrace
```
jmaupetit, line 26: IMO the `--prefix` is not required: we do not need to split our dependencies in different places.

tasks/get_objects_for_app.yml
```diff
@@ -6,6 +6,16 @@
     templates: "{{ app | json_query('services[*].templates[]') | list }}"
   tags: deploy
 
+- name: Set OpenShift datadog objects to manage
+  set_fact:
+    images: "{{ templates | map('regex_search', '.*/is.yml.j2$') | select('string') | list }}"
+    builds: "{{ templates | map('regex_search', '.*/bc.yml.j2$') | select('string') | list }}"
```
jmaupetit, line 12: Those objects should be generic (imagestreams and buidconfigs), not only related to datadog APM. Hence you should move them in the next bloc.

tasks/get_objects_for_app.yml
```diff
@@ -6,6 +6,16 @@
...
+  when: ddtrace
```
jmaupetit, line 13: We always want to collect those objects not only when `ddtrace` is true. They are already conditionally created _via_ their templates.

tasks/get_objects_for_app.yml
```diff
@@ -19,6 +29,14 @@
     - job
     - route
 
+- name: Display OpenShift's build for this app
+  debug: msg="{{ builds | to_nice_yaml}}"
+  when: builds is defined
+  when: ddtrace
```
jmaupetit, line 35: Remove this condition.

tasks/manage_app.yml
```diff
@@ -7,6 +7,19 @@
...
+    definition: "{{ lookup('template', item) | from_yaml }}"
+    state: "{{ deployment_state | default('present') }}"
+  with_items:
+    - "{{ images }}"
+    - "{{ builds }}"
+  when: ddtrace
+  tags:
+    - deploy
+    - image
+    - build
```
jmaupetit, line 21: Same idea here: you should create ISs and BCs as generic objects. This should not be datadog-specific.

apps/richie/templates/app/dc.yml.j2
sampaccoud, line 56: it should come from a variable?

group_vars/all/main.yml
```diff
@@ -62,6 +62,12 @@ job_stamp: null
 # docs/developer_guide/secrets.md
 secret_id: "1.0.0"
 
+
+# ddtrace is set to true for enable the APM in all apps where the templates
+# include the possibility to have a ddtrace APM agent
+# defaul is set to False
+ddtrace: false
```
jmaupetit, line 69: Should we name it `is_apm_enabled` instead of `ddtrace` which is less explicit?

group_vars/all/main.yml
```diff
@@ -62,6 +62,12 @@ job_stamp: null
 # docs/developer_guide/secrets.md
 secret_id: "1.0.0"
 
+
+# ddtrace is set to true for enable the APM in all apps where the templates
+# include the possibility to have a ddtrace APM agent
+# defaul is set to False
```
jmaupetit, line 68: I would write:

```yaml
# Set xxx to "true" to enable APM in all apps where defined templates
# include this feature (default: "false")
```

jmaupetit (commented):
That's a smart idea :clap:

apps/richie/templates/app/dc.yml.j2
sampaccoud: We discussed this with @rouja and propose to always point to an image stream.
Always having an image stream is an optimization we want because it stops downloading all the docker image layers on each deployment.

The syntax is more or less:
```
template:
    spec:
        containers:
        - name: "richie"
          image: ""
triggers:
    - type: ConfigChange
    - type: ImageChange
      imageChangeParams:
          automatic: true
          containerNames:
          - "richie"
          from:
            kind: ImageStreamTag
            name: "{{ richie_image_name }}:{{ richie_image_tag }}"
```

Then in the image stream, we build the image we need for each project:
- adding monitoring (or not)
- for edxapp: adding a theme (or not)
- etc.

jmaupetit: That's a wonderful feature. I'll declare a new issue for this.

sampaccoud: @jmaupetit There is already a PR for this ;-) https://github.com/openfun/arnold/pull/105

jmaupetit: :thinking: yes, but it only scopes edxapp right?

sampaccoud: True! But it's a functional POC whereas my comment above was a bit vague.

PR #105, opened 2018-08-27
WIP: Add a custom theme to edxapp LMS

sampaccoud:
## Purpose

Adding a custom theme to an Open edX site must be done upon deployment and not in the common Docker image because it is specific to each customer.

**edit**: this PR is depends on #111 

## Proposal

We propose to use OpenShift's buildconfig objects to automate such builds. The same mechanism can be used to add monitoring or other build steps that are specific either to our way to deploy things to OpenShift or to a customer (like the theme).

It is a good idea to use a common image stream for the CMS and the LMS even if no customization is made because it allows keeping a copy of the image in our local registry and avoid downloading it
from DockerHub upon each deployment.

- [x] Add a build config step and an image stream
- [ ] Upgrade edxapp docker image version to gingko.1-1.0.7
- [ ] Remove the DEFAULT_SITE_THEME env var (see https://github.com/openfun/arnold/pull/105#discussion_r213377200)
- [ ] Add a private SSH key as a sshauth secret to connect to the theme as a private repository in Gitlab

### DIFF
--- a/apps/edxapp/templates/lms/bc.yml.j2
+++ b/apps/edxapp/templates/lms/bc.yml.j2
@@ -1,0 +1,30 @@
+apiVersion: "v1"
+kind: "BuildConfig"
+metadata:
+  name: "edxapp-lms"
+  labels:
+    app: "edxapp"
+    service: "lms"
+spec:
+  strategy:
+    dockerStrategy:
+  source:
+    {% if edxapp_theme_url is defined and edxapp_theme_url %}
+    git:
+      uri: "{{ edxapp_theme_url }}"
+      ref: "{{ edxapp_theme_tag | default("master") }}"
+    sourceSecret:
+      name: "edxapp_theme_ssh_key"
+    {% endif %}
+    dockerfile: |-
+      FROM {{ edxapp_image_name }}:{{ edxapp_image_tag }}
+      USER 0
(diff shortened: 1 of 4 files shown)


apps/edxapp/templates/lms/bc.yml.j2
```diff
@@ -0,0 +1,30 @@
...
+      ref: "{{ edxapp_theme_tag | default("master") }}"
+    sourceSecret:
+      name: "edxapp_theme_ssh_key"
+    {% endif %}
+    dockerfile: |-
+      FROM {{ edxapp_image_name }}:{{ edxapp_image_tag }}
+      USER 0
+      {% if edxapp_theme_url is defined and edxapp_theme_url %}
+      COPY . /edx/app/edxapp/edx-platform/themes/custom-theme
+      NO_PREREQ_INSTALL=1 DEFAULT_SITE_THEME="custom-theme" paver update_assets --settings={{ edxapp_build_settings }} --skip-collect
```
sampaccoud, line 24: @jmaupetit in fact, I think it might not be required to set DEFAULT_SITE_THEME since "update_assets" builds all the themes in the themes directory... 
Maybe try removing it as it would be simpler... and we wouldn't need to modify the build settings in fun-platform like I did...

PR #109, opened 2018-08-29
✨(script:dev) allow to set (or guess) local IP address

jmaupetit:
## Purpose

The `bin/dev` script is only compatible with GNU/Linux and suppose that the first IP address listed _via_ the `hostname -I` command matches the interface that has access to the internet. This is potentially false, we cannot rely on such expectation. 

## Proposal

In the `bin/dev` script, we now test all IP addresses of all interfaces and set the first one with access to the internet as the `OPENSHIFT_DOMAIN`.

Plus, we now allow to set the `OPENSHIFT_DOMAIN` IP address as a script argument, hence, MacOSX users can now use the `bin/dev` script as follows:

```bash
$ bin/dev 192.168.1.10
```

(with `192.168.1.10` as their local IP address that has access to the internet)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,10 +126,11 @@ $ oc login --insecure-skip-tls-verify=true -u developer -p developer "${K8S_AUTH
 ```
 
 As we are gentle people, we also provide a shortcut to automate running a new
-cluster and login to it (GNU/Linux only for now):
+cluster and login to it:
 
-```
-$ bin/dev
+```bash
+# Substitute 192.168.1.10 with your local IP that has access to the internet
+$ bin/dev 192.168.1.10
 ```
 
 Please, note that **you do not need to run** `bin/dev` if you have already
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -126,10 +126,11 @@ $ oc login --insecure-skip-tls-verify=true -u developer -p developer "${K8S_AUTH
 ```
 
 As we are gentle people, we also provide a shortcut to automate running a new
-cluster and login to it (GNU/Linux only for now):
+cluster and login to it:
 
-```
-$ bin/dev
+```bash
+# Substitute 192.168.1.10 with your local IP that has access to the internet
```
sampaccoud, line 132: only for MAC users, linux users can just run bin/dev...

sampaccoud (approved):

README.md
jmaupetit: Well, as this feature is still experimental, I prefer a "solution that works everywhere:tm:" for a quick start and not having to deal with particular cases/systems.

jmaupetit: But this will be detailed in the documentation.

Issue #106, opened 2018-08-28
Consider moving all applications' docker images to imageStreams


jmaupetit:
## Purpose

As discussed in [1], using imageStream objects instead of raw images allows substantial optimizations to decrease network bandwidth (while pulling docker images), add monitoring to our applications, or customize our builds (_e.g._ add a theme for edxapp, etc.).

## Proposal

A first POC for edxapp has been implemented in #105, this should be generalized to all distributed apps (at least richie and hello at the time of writing).

## References

[1] https://github.com/openfun/arnold/pull/101#discussion_r213030738

jmaupetit:
Already fixed.

PR #111, opened 2018-08-30
Add support for ImageSteam and BuildConfig objects

jmaupetit:
## Purpose

Until now, BuildConfiguration and ImageStream objects were not required. Now that we need to add APM and customize images on a per-client basis, using the aforementioned objects seems the more relevant solution.

_nota bene_: 

* this PR is closely related to #101 and #105 as they both require support for those objects.
* this PR also address #106 

## Proposal

- [x] add basic/generic support for `ImageStream` and `BuildConfig` objects
- [x] add templates to use an IS/BC for `edxapp` (lms + cms)
- [x] add templates to use an IS/BC for `richie` (app)

### DIFF
--- a/apps/edxapp/templates/cms/_dc_base.yml.j2
+++ b/apps/edxapp/templates/cms/_dc_base.yml.j2
@@ -57,7 +57,9 @@ spec:
           value: {{ service_variant }}
         - name: DJANGO_SETTINGS_MODULE
           value: {{ service_variant }}.envs.fun.docker_run
-        image: "{{ edxapp_image_name }}:{{ edxapp_image_tag }}"
+        # We point to a local registry image build for this "live" image (see
+        # ImageStream and BuildConfig templates)
+        image: "{{ internal_docker_registry }}/{{ project_name }}/edxapp-{{ service_variant }}:{{ edxapp_image_tag }}-live"
         imagePullPolicy: IfNotPresent
         volumeMounts:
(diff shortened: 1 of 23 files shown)


apps/edxapp/templates/cms/_dc_base.yml.j2
```diff
@@ -57,7 +57,9 @@ spec:
           value: {{ service_variant }}
         - name: DJANGO_SETTINGS_MODULE
           value: {{ service_variant }}.envs.fun.docker_run
-        image: "{{ edxapp_image_name }}:{{ edxapp_image_tag }}"
+        # We point to a local registry image build for this "live" image (see
+        # ImageStream and BuildConfig templates)
+        image: "{{ internal_docker_registry }}/{{ project_name }}/edxapp-{{ service_variant }}:{{ edxapp_image_tag }}-live"
```
sampaccoud, line 62: This is complicated and not necessary if you do as in my POC. Set the image to an empty string and use triggers to set it?

sampaccoud (approved):

apps/edxapp/templates/cms/_dc_base.yml.j2
jmaupetit: No, because we want more control on our deployments. Adding a trigger on this DCs leads to some uncertainties I don't want to struggle upon :wink:

PR #115, opened 2018-09-05
WIP: Add datadog APM support

jmaupetit:
## Purpose

This is a follow up of #101 considering recent changes concerning BuildConfig and ImageStream. We've decided to add APM to our apps by adding new layers to base images (_via_ service BuildConfig).

## Proposal

- [x] add APM related variables to make them configurable (env_type, customers)
- [ ] integrate APM for `richie`
- [ ] integrate APM for `edxapp`

Closes #101

### DIFF
--- a/apps/richie/templates/app/bc.yml.j2
+++ b/apps/richie/templates/app/bc.yml.j2
@@ -29,7 +29,12 @@ spec:
   source:
     dockerfile: |-
       FROM {{ richie_image_name }}:{{ richie_image_tag }}
-      # Add new statements here
+      USER 0
+      {% if is_apm_enabled -%}
+      RUN pip install ddtrace
+      CMD ddtrace-run gunicorn -c /usr/local/etc/gunicorn/richie.py richie.wsgi:application
+      {% endif -%}
+      USER 10000
   triggers:
     - type: "ConfigChange"
   output:
--- a/apps/richie/templates/app/dc.yml.j2
+++ b/apps/richie/templates/app/dc.yml.j2
@@ -40,6 +40,18 @@ spec:
               value: "{{ richie_host }}"
             - name: ES_CLIENT
               value: "richie-{{ richie_elasticsearch_host }}-{{ deployment_stamp }}"
+            {% if is_apm_enabled -%}
+            - name: DATADOG_TRACE_AGENT_HOSTNAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_HOST
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_PORT
+              value: "{{ apm_agent_port }}"
+            {% endif -%}
           envFrom:
             - secretRef:
                 name: richie-{{ secret_id }}
--- a/apps/richie/templates/app/dc.yml.j2
+++ b/apps/richie/templates/app/dc.yml.j2
@@ -40,6 +40,7 @@ spec:
               value: "{{ richie_host }}"
             - name: ES_CLIENT
               value: "richie-{{ richie_elasticsearch_host }}-{{ deployment_stamp }}"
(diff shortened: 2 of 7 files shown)


apps/richie/templates/app/bc.yml.j2
```diff
@@ -29,7 +29,12 @@ spec:
   source:
     dockerfile: |-
       FROM {{ richie_image_name }}:{{ richie_image_tag }}
-      # Add new statements here
+      USER 0
+      {% if is_apm_enabled -%}
```
sampaccoud, line 33: add a comment line because this Dockerfile extract will receive several separate augmentations?

group_vars/all/main.yml
```diff
@@ -65,6 +65,12 @@ secret_id: "1.0.0"
 # OpenShift's internal docker registry server
 internal_docker_registry: "docker-registry.default.svc:5000"
 
+# Set is_apm_enabled to "true" to enable APM in all apps where defined templates
+# include this feature (default: "false")
+is_apm_enabled: false
+# APM agent port defaults to Datadog agent service port
+apm_agent_port: 8126
```
sampaccoud, line 72: Would be nice if we could omit the port when it is disabled.

jmaupetit: This is the default (main) value. When it's disabled, you don't need to declare this variable at all.

apps/richie/templates/app/bc.yml.j2
```diff
@@ -29,7 +29,12 @@ spec:
   source:
     dockerfile: |-
       FROM {{ richie_image_name }}:{{ richie_image_tag }}
-      # Add new statements here
+      USER 0
```
cchaudier, line 32: Why creating a BC who do nothing if we have no apm ?

apps/richie/templates/app/bc.yml.j2
```diff
@@ -29,7 +29,12 @@ spec:
...
+      {% if is_apm_enabled -%}
+      RUN pip install ddtrace
```
cchaudier, line 34: This code is only for datadog so why the variable is named is_apm_enabled ?
If we use an other APM we need a new name.
I propose is_apm_datadog_enabled, if we add newrelic whe can use is_apm_newrelic_enabled.

Or we can specify the APM name on the variable, so s_apm_enabled can ave the value (false, datadog, newrelic, etc.)

apps/richie/templates/app/dc.yml.j2
```diff
@@ -40,6 +40,18 @@ spec:
...
+            - name: DATADOG_TRACE_AGENT_HOSTNAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_HOST
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: DD_AGENT_SERVICE_PORT
+              value: "{{ apm_agent_port }}"
```
cchaudier, line 54: Can you add a default value ?

group_vars/all/main.yml
cchaudier: See my other message, you can add a default value on the BC template

cchaudier (commented):
Hello, I've juste somme comments.

jmaupetit:
@cchaudier `is_apm_enabled` is a feature flag that is proposed by Arnold, allowing you to enable/disable APM for your app services, if and only if you have implemented such feature in your application's templates.

Hence, Arnold does not have to be aware of the APM tool you choose. It's the app developer responsibility to implement such feature with datadog or any other tool/service.

apps/richie/templates/app/bc.yml.j2
jmaupetit: To take advantage of the cluster's docker repository and speed up image pulling. We only create a IS/BC for our public images that will:

1. will be updated frequently
2. requires further customization given a `customer` and an `env_type`

PR #120, opened 2018-09-10
Add a mail catcher app

sampaccoud:
## Purpose

In non-production environments, we don't want  to send real emails but we want to easily consult the emails sent to anyone.

## Proposal

The solution for this is to use a mail catcher. There are several solutions available and we chose https://github.com/sj26/mailcatcher because it is easy, good enough and distributed as a Docker image.

- [x] Add a separate mailcatcher app not subject to blue/green deployments
- [x] Configure edxapp to send its emails to mailcatcher

### DIFF
--- a/apps/mailcatcher/templates/app/dc.yml.j2
+++ b/apps/mailcatcher/templates/app/dc.yml.j2
@@ -1,0 +1,29 @@
+apiVersion: v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    app: mailcatcher
+    service: app
+    version: "{{ mailcatcher_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: "mailcatcher-app"
+  namespace: "{{ project_name }}"
+spec:
+  replicas: 1  # number of pods we want
+  template:
+    metadata:
+      labels:
+        app: mailcatcher
+        service: app
+        version: "{{ mailcatcher_image_tag }}"
+        deployment_stamp: "{{ deployment_stamp }}"
+        deploymentconfig: "mailcatcher-app"
+    spec:
+      containers:
+      - name: mailcatcher
+        image: "{{ mailcatcher_image_name }}:{{ mailcatcher_image_tag }}"
+        ports:
+        - containerPort: 1025
+          protocol: TCP
+        - containerPort: 1080
+          protocol: TCP
--- a/apps/mailcatcher/templates/app/svc.yml.j2
+++ b/apps/mailcatcher/templates/app/svc.yml.j2
@@ -6,7 +6,7 @@ metadata:
     service: app
     version: "{{ mailcatcher_image_tag }}"
     deployment_stamp: "{{ deployment_stamp }}"
-  name: "{{ mailcatcher_app_host }}"
+  name: "mailcatcher"
   namespace: "{{ project_name }}"
 spec:
   ports:
--- a/apps/mailcatcher/templates/app/svc.yml.j2
+++ b/apps/mailcatcher/templates/app/svc.yml.j2
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: mailcatcher
+    service: app
+    version: "{{ mailcatcher_image_tag }}"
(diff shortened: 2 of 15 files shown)


apps/mailcatcher/templates/app/dc.yml.j2
```diff
@@ -0,0 +1,33 @@
...
+    version: "{{ mailcatcher_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: "mailcatcher-app"
+  namespace: "{{ project_name }}"
+spec:
+  replicas: 1  # number of pods we want
+  # When upgrading, we don't need OpenShift to run several pods mounted on the same volume
+  # Instead we need to make sure that the existing pod is switched off before the new one starts.
+  strategy:
+    type: Recreate
```
jmaupetit: Since you don't have any volume, this is not required.

group_vars/customer/patient0/development/main.yml
```diff
@@ -1,6 +1,7 @@
 # Variables specific to the patient0 customer in development env_type
 
 apps:
+  - name: mailcatcher
```
jmaupetit, line 4: Can you also add it to `env_types` where it is required (_i.e._ `group_vars/env_type/*`)?

sampaccoud: Done

jmaupetit: I only see CI and development, I thought we would use it everywhere but in production?

apps/mailcatcher/templates/app/svc.yml.j2
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: mailcatcher
+    service: app
+    version: "{{ mailcatcher_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: "{{ mailcatcher_app_host }}"
```
jmaupetit: I would call it `mailcatcher` and remove this confusing variable (`mailcatcher_app_host` _versus_ `mailcatcher_host`).

apps/mailcatcher/vars/all/main.yml
```diff
@@ -0,0 +1,11 @@
+# Application default configuration
+
+# -- route
+mailcatcher_host: "mailcatcher.{{ project_name}}.{{ domain_name }}"
+
+# -- mailcatcher
+mailcatcher_image_name: "sj26/mailcatcher"
+mailcatcher_image_tag: "latest"
+mailcatcher_app_host: "mailcatcher"
```
jmaupetit: As said, I would remove this var.

jmaupetit (commented):
This is odd that we cannot see you new CI step (mailcatcher bootstrap)?

sampaccoud:
@jmaupetit not weird: I forgot to add it to the list of jobs...

jmaupetit:
I should have seen this... well, now it works! :tada:

group_vars/customer/patient0/development/main.yml
sampaccoud: @jmaupetit there was no "apps" parameter to add it to so I thought these environments are not yet used anyway.

jmaupetit: Ok fine. In this case, it means that every app is considered as deployable.

apps/mailcatcher/templates/app/svc.yml.j2
sampaccoud: but there is already a "mailcatcher_host" here https://github.com/openfun/arnold/pull/120/files#diff-1d6d10331c2861156e52d71455fd392dR16 so I followed naming from the "redis" app.

jmaupetit: Sorry I did not mean to call your variable `mailcatcher`, but rather hard code the name:

```yaml
name: "mailcatcher"
```

sampaccoud: oh yeah you're right!

jmaupetit (approved):
Ship it!

PR #123, opened 2018-09-11
⚡️(acme) Make openshift-acme a core application

jmaupetit:
## Purpose

`openshift-acme` should be considered as a core application as it will always get deployed for non-development projects. Core applications should get handled similarly to other apps.

## Proposal

Instead of downloading templates from a github repository, we've "vendored" them in the repository, and we've updated the `acme` application's deployment.


core_apps/acme/templates/app/dpl.yml.j2
```diff
@@ -0,0 +1,49 @@
+kind: Deployment
```
sampaccoud, line 1: I must be missing something but dpl.yml should be dc.yml?

core_apps/acme/templates/app/dpl.yml.j2
```diff
@@ -0,0 +1,49 @@
...
+apiVersion: apps/v1beta1
+metadata:
+  name: openshift-acme
+  labels:
+    app: openshift-acme
+  annotations:
+    image.alpha.openshift.io/triggers: '[{"from":{"kind":"ImageStreamTag","name":"openshift-acme:latest"},"fieldPath":"spec.template.spec.containers[0].image"}]'
+    image.openshift.io/triggers: '[{"from":{"kind":"ImageStreamTag","name":"openshift-acme:latest"},"fieldPath":"spec.template.spec.containers[?(@.name==\"openshift-acme\")].image"}]'
```
sampaccoud, line 9: I find this hard to understand :-(

core_apps/acme/templates/app/dpl.yml.j2
jmaupetit: Well, it's a `Deployment`, not a `DeploymentConfig`, so...

core_apps/acme/templates/app/dpl.yml.j2
jmaupetit: In fact, this is one of the original `openshift-acme` templates we used that I vendored... But yes: I must admin that is not straightforward.

sampaccoud (approved):

PR #126, opened 2018-09-13
Improve BuildConfig triggering

jmaupetit:
## Purpose

Instead of building images at every deployment, we need to make it a separated process to avoid using resources when it's not necessary and potentially harmful as it slows down deployments.

## Proposal

- [x] remove builds from deployment
- [x] add a `build_images.yml` playbook
- [x] ~add docker hub webhooks to trigger a new build when a new image has been published~ \*

\* After thought: this might be a bad idea as it would require that we create one webhook per image per project (_i.e._ a `customer` in an `env_type`), and each new image publication would trigger a lot of builds that we are not sure to be able to support.

### DIFF
--- a/tasks/create_app_builds.yml
+++ b/tasks/create_app_builds.yml
@@ -1,0 +1,15 @@
+---
+# Create BuildConfigs for an app
+
+- name: Print app name
+  debug: msg="App name {{ app.name }}"
+  tags: build
+
+- name: Make sure build configs exist
+  openshift_raw:
+    definition: "{{ lookup('template', item) | from_yaml }}"
+    state: present
+    force: true
+  with_items: "{{ builds }}"
+  when: builds is defined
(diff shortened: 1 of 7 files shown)


tasks/create_app_builds.yml
```diff
@@ -0,0 +1,15 @@
+---
+# Create BuildConfigs for an app
+
+- name: Print app name
```
sampaccoud, line 4: But in fact we need a different build for each customer and it needs the address and ssh key to the private repo with the theme for this customer?

jmaupetit: This will make a build for a project (aka namespace), hence a customer in an environment for us. Dealing with private repositories is another feature IMO. WDYT?

sampaccoud (approved):

PR #130, opened 2018-09-20
Improve secrets management

jmaupetit:
## Purpose

Secrets generation is for now limited to a generic template that handles key:values that can be injected as environment variables to a running pod. We are not able to define other type of secrets than Opaque ones, and we are restricted to one secret per application. We think there is room for improvement.

## Proposal

- [x] allow to define and create secrets template in service templates for an app
- [x] add new templates for `richie`'s secrets
- [x] add new templates for `edxapp`'s secrets


apps/edxapp/templates/mongodb/secret.yml.j2
```diff
@@ -0,0 +1,13 @@
...
+kind: Secret
+metadata:
+  labels:
+    app: edxapp
+    service: mongodb
+  name: "{{ edxapp_mongodb_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  MONGODB_ADMIN_PASSWORD: "{{ MONGODB_ADMIN_PASSWORD | default('password') | b64encode }}"
+  MONGODB_USER: "{{ MONGODB_USER | default('fun') | b64encode }}"
```
sampaccoud: Should we call it "edxapp_user"?

apps/richie/templates/app/secret.yml.j2
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  labels:
+    app: richie
+    service: richie
+  name: "{{ richie_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  DJANGO_SECRET_KEY: "{{ DJANGO_SECRET_KEY | default('supersecret') | b64encode }}"
+  DJANGO_SENTRY_DSN: "{{ DJANGO_SENTRY_DSN | default('fakedsn') | b64encode }}"
+  POSTGRES_USER: "{{ POSTGRESQL_USER | default('fun') | b64encode }}"
```
sampaccoud: ditto

apps/richie/templates/postgresql/secret.yml.j2
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: richie
+    service: postgresql
+  name: "{{ richie_postgresql_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  POSTGRESQL_USER: "{{ POSTGRESQL_USER | default('fun') | b64encode }}"
```
sampaccoud: ditto

tasks/get_objects_for_app.yml
```diff
@@ -46,6 +46,7 @@
 - name: Display OpenShift's builds for this app
```
sampaccoud, line 46: in commit message: s/should'nt/shouldn't

sampaccoud (approved):

PR #131, opened 2018-09-28
Minor fixes to smooth devops experience

sampaccoud:
## Purpose

I and an external user bumped into small glitches while trying to deploy Open edX projects. Let's smooth things out.

## Proposal

- [x] We only want to add the project display name with the "env_type" between brackets for the name of the project in OpenShift. In edX, the PLATFORM_NAME is used for example in buttons and should therefore be a short display name.
- [x] Add a bin script to build images as it was recently removed from the deploy process,
- [x] Rename the `preproduction` environment to preprod for coherence with fun-platform.


group_vars/customer/campus/main.yml
```diff
@@ -1,6 +1,6 @@
 # Variables specific to the campus customer
-project_display_name: "FUN Campus ({{ env_type }})"
-project_description: "SPOC for universities ({{ env_type }})"
+project_display_name: "FUN Campus"
+project_description: "SPOC for universities"
```
jmaupetit: IMO there is a confusion between the OpenShift and the application layer. In an OpenShift context I want to get the `env_type`. I think, this should be fixed in the application layer (set the `SITE_NAME` to `{{ customer }}` for example). Do you read me?

bin/build_images
```diff
@@ -0,0 +1,11 @@
...
+
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+# Set OpenShift's environment (you'll need to be logged in to an OpenShift
+# server)
+_set_openshift_env
+
+# Run ansible-playbook
+_docker_run ansible-playbook build_images.yml "$@"
```
jmaupetit, line 11: :medal_sports: I've missed this one.

jmaupetit: FYI you are not forced to use the whole docker command, I used the `bin/ansible-playbook` helper.

jmaupetit (approved):

PR #143, opened 2018-10-22
Add forum

jmaupetit:
## Purpose

Edx comments service is a core part of the edx-platform companions. Once properly dockerized, the application requires to be deployable thanks to Arnold.

## Proposal

- [x] add base OpenShift objects for the forum application
- [x] add required dependencies (`mongodb`, `elasticsearch`)
- [x] add tests in the CI
- [x] add ES init index job

## Related projects

See https://github.com/openfun/openedx-comments-docker


apps/forum/templates/app/dc.yml.j2
```diff
@@ -0,0 +1,37 @@
...
+          envFrom:
+            - secretRef:
+                name: "{{ forum_secret_name }}"
+          env:
+            - name: SEARCH_SERVER
+              value: "http://forum-elasticsearch-{{ deployment_stamp }}:{{ forum_elasticsearch_port }}"
+          command:
+            - "/bin/bash"
+            - "-c"
+            - MONGOHQ_URL=mongodb://${MONGODB_USER}:${MONGODB_PASSWORD}@forum-mongo-{{ deployment_stamp }}:{{ forum_mongodb_port }}/${MONGODB_DATABASE} bundle exec puma
```
sampaccoud, line 40: why not include it in the secrets that are mounted as env variables?

sampaccoud (approved):

apps/forum/templates/app/dc.yml.j2
jmaupetit: Because of the deployment stamp that is part of the service host name: it changes at every deployment, and secrets generation is separated from those deployments (created during project initialization).

PR #149, opened 2018-11-08
Fix nginx configurations to work behind the TLS termination proxy

sampaccoud:
## Purpose

`Nginx` acts as a proxy and was overriding the value of the `X-Forwarded-` headers. The problem is that our applications are running in `OpenShift` behind a `HAProxy` that handles the TLS termination. This proxy has already set the `X-Forwarded-` headers with the values received from the client, so we should respect them and not override them with the values in the request hitting `Ǹginx` because they come from the `HAProxy`, not the client.

## Proposal

Stop overriding the `X-Forwarded-` headers in `Nginx`.


jmaupetit (approved):
LGTM

PR #150, opened 2018-11-08
⚡️(dc) move imagePullPolicy to Always

jmaupetit:
## Purpose

By relying on the `ifNotPresent` image pull policy, we aren't sure which image version will be deployed if we've just built and pushed target image (with the same name and tag).

## Proposal

- [x] Switch to the `Always` image pull policy to ensure local image is up-to-date

Note: We've also fixed indentation for some deployment configurations.


sampaccoud (approved):

lunika (approved):
Good moove. I always do that too.

PR #151, opened 2018-11-13
✨(security) add http basic auth for app services

jmaupetit:
## Purpose

In some cases, we must be able to protect a publicly deployed service behind an HTTP Basic Auth, _e.g._ to prevent search engines bots crawling or restrict access to a confidential application instance.

## Proposal

- [x] automate htpasswd file generation for apps
- [x] integrate HTTP Basic Auth for `richie`
- [x] integrate HTTP Basic Auth for `marsha`
- [x] integrate HTTP Basic Auth for `edxapp`

### DIFF
--- a/apps/richie/templates/nginx/secret.yml.j2
+++ b/apps/richie/templates/nginx/secret.yml.j2
@@ -1,0 +1,14 @@
+{% if activate_http_basic_auth %}
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: richie
+    service: "nginx"
+  name: "{{ richie_nginx_htpasswd_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  # nota bene: the {{ app.name }}_htpasswd variable is set in
+  # tasks/get_vault_for_app.yml tasks list only if the pointed file exists
+  "{{ http_basic_auth_user_file | basename }}": "{{ lookup('file', richie_htpasswd) | b64encode }}"
+{% endif %}
--- a/tasks/get_vault_for_app.yml
+++ b/tasks/get_vault_for_app.yml
@@ -3,7 +3,7 @@
 # group_vars for an environent and a customer.
 - name: Get expected vault file for application
   stat:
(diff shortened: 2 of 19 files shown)


tasks/get_vault_for_app.yml
```diff
@@ -19,3 +19,20 @@
...
+- name: Test htpasswd target file for app
+  stat:
+    path: "{{ app_htpasswd_file_path }}"
+  register: app_htpasswd_file
+  tags: htpasswd
+
+- name: Set application htpasswd variable
+  set_fact:
+    "{{ app.name }}_htpasswd": "{{ app_htpasswd_file.stat.path }}"
+  when: activate_http_basic_auth and app_htpasswd_file.stat.exists and app_htpasswd_file.stat.isreg
```
sampaccoud, line 39: can you comment what isreg is doing checking? here or in the previous task?

apps/richie/templates/nginx/secret.yml.j2
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: richie
+    service: "nginx"
+  name: "{{ richie_nginx_htpasswd_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  "{{ http_basic_auth_user_file | basename }}": "{{ lookup('file', richie_htpasswd) | b64encode }}"
```
sampaccoud, line 13: took me some time to understand wher the "richie_htpasswd" variable comes from (same for marsha and edxapp), maybe add a comment with link to the line where it's done in `get_vault_for_app`.

sampaccoud (approved):

sampaccoud:
@jmaupetit awesome feature, and nicely done! Just added comments where I found it hard to understand if you want to add comments.

apps/richie/templates/nginx/secret.yml.j2
jmaupetit: Yeah, that's a tricky part. Maybe I should make it simpler.

PR #152, opened 2018-11-15
Add missing secret update state shared secrets to marsha

sampaccoud:
## Purpose

This secret is necessary for the AWS lambda `lambda-update-state` to call Django and update the status of the video. It was forgotten because it was not [properly documented in `Marsha`](https://github.com/openfun/marsha/blob/master/docs/env.md#django_update_state_shared_secrets) until recently.

We need to add this secret in `Arnold` before we can set its value in our vaults in `Matsuo`.

## Proposal

Add it the `DJANGO_UPDATE_STATE_SHARED_SECRETS` key to secrets in `Marsha`.


jmaupetit (approved):
LGTM

jmaupetit (approved):
:clap:

Issue #153, opened 2018-11-15
Better normalize and enforce our database names


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We need to normalize and enforce database names because MySQL limits db names to **16 characters** by default.

**Describe the solution you'd like**
Name databases with a pattern: `{{ env_type }}-{{ customer }}-{{ app.name }}[{{ extra }}]`
where:

- `customer` is max 6 characters
- `env_type` is 1 character (because we are limited to 16 characters, we do not use 2 characters for this information):
    * `s`: staging
    * `t`: pre-production (after s... try/test...)
    * `p`: production
- `app.name` is limited to 6 characters: _e.g._ `richie`, `marsha`, `edxapp`, etc.
- `extra` is 1 character that can be added if we have several databases for a given customer/env_type/database.

This normalization will allow us to generate interface documents to automate the creation of all our databases. We propose the following format for this document:

```yaml
postgresql:
  9_10:
    # Richie staging for customer foo
    - name: s_foo_richie
      user: s_foo_richie
      password: 123456
    # Richie production for customer foo
    - name: p_foo_richie
      user: p_foo_richie
      password: 123456
    # Richie staging for customer bar
    - name: p_bar_richie
      user: p_bar_richie
      password: 123456
    # Marsha preproduction for customer bar
    - name: t_bar_marsha
      user: t_bar_marsha
      password: 123456
mysql:
  5_6:
    # Mysql production for customer foo
    - name: p_foo_edx
      user: p_foo_edx
      password: 123456
mongodb:
  3_0:
    # Mongodb production for customer foo on Mongodb 3.0
    - name: p_foo_edx
      user: p_foo_edx
      password: 123456
  3_2:
    # Mongodb production for customer bar on Mongodb 3.2
    - name: p_bar_edx
      user: p_bar_edx
      password: 123456
```

We must:

- [ ] modify our template and variables,
- [ ] add checks to enforce this pattern,
- [ ] generate interface vault files from which we can automate creation of these databases.

**Describe alternatives you've considered**
We hesitated with a pattern `{{ customer ]}-{{ env_type }}-{{ app}}` but:

- the project_name is  `{{ env_type }}-{{ customer }}`,
- since we shorten the `env_type` name, I would argue that `p-client-marsha` is nicer than `client-p-marsha`,
- `env_type` is the first differentiator (e.g production is usually a whole separate infrastructure).

jmaupetit:
[...]

PR #161, opened 2018-11-21
Check customer name length is not higher than 6 characters

lunika:
## Purpose

In #153 we decided to not have customer name length higher than 6 characters. This rule is here to be sure to have a mysql username not higher than 16 characters.

## Proposal

before running every playbooks we should check than the customer name respect the rule described before.
To do that, we created a pre_task imported in all existing playbooks.


pre_tasks/customer_name_verification.yml
```diff
@@ -0,0 +1,4 @@
+- name: Check if customer name length is 6 characters length max
```
sampaccoud: Check that customer name is maximum 6 characters long?

sampaccoud:
@lunika if you rebase on master the failing test-plugins will resolve.

sampaccoud:
@lunika maybe add explanation somewhere why we limit to 6 characters? doc or just a comment in the code?

lunika:
> @lunika maybe add explanation somewhere why we limit to 6 characters? doc or just a comment in the code?

Done. I had an explanation in the documentation.

pre_tasks/customer_name_verification.yml
```diff
@@ -0,0 +1,4 @@
+- name: Check that customer name is maximum 6 characters long
+  fail:
+    msg: "Customer name length can not be higher than 6 characters. Current length is {{ customer | length }}."
+  when: customer | length > 6
```
jmaupetit: I would suggest to rename this pre-task: `validate_customer_name.yml`

docs/developer_guide/concepts.md
```diff
@@ -36,8 +36,10 @@ where:
 
 * `environment`: the name of the environment (development, feature, staging,
   preprod or production)
-* `customer`: the one-word name of the customer site (fun, campus, corporate,
-  ademe, etc.)
+* `customer`: the one-word name of the customer site (fun, campus, ademe, 
+etc.). Max length for the customer name is 6. Because mysql username has a 
+length constraint we can not have a customer name higher than 6 characters
+long.
```
jmaupetit: Suggestion:

```
* `customer`: the one-word name of the customer site (_e.g._ `ucsd`, `hvrd`,
  `uca`). We enforce this name to at most 6 characters, because it is used in a
  strict database user & name pattern that must also include the environment and
  targeted application name within 16 characters (yes MySQL we are looking at
  you 😉).
```

jmaupetit (commented):
That's a great piece of work @lunika! Thx!

lunika:
modification made @sampaccoud @jmaupetit

jmaupetit (approved):

lunika:
@jmaupetit what do you think about moving `validate_customer_name.yml` file to `tasks folder ? It's a task, you can use it where you want.

jmaupetit:
Yup. Good idea.

PR #162, opened 2018-11-22
🚸(marsha) redirect http requests to https

sampaccoud:
## Purpose

On [marsha](https://github.com/openfun/marsha), we want all insecure requests to be automatically redirected to https.

## Proposal

I added the same rule as what we have in the route definitions of our other apps.


jmaupetit (approved):
:+1:

PR #164, opened 2018-11-26
🔥(oc) remove oc client in arnold's image

### LINKED ISSUE #107: Consider removing oc client from Arnold's docker image
## Purpose

Installing `oc` in arnold's docker image seemed a pretty good idea, as docker would have been the only requirement for Arnold. Unfortunately, oc stores a lot of data in the user space, and, thus we never used it.

## Proposal

- [ ] Remove `oc` installation from Arnold's Dockerfile
- [ ] Update the documentation (at least the README)

lunika:
## Purpose

Installing oc in arnold's docker image seemed a pretty good idea, as docker would have been the only requirement for Arnold. Unfortunately, oc stores a lot of data in the user space, and, thus we never used it.

Fixes #107 

## Proposal

Remove the oc client installation in Dockerfile and its usage in `entrypoint`. We also need to share the `$HOME/.kube` folder in the container to share context and config.

### DIFF
--- a/bin/ci-bootstrap
+++ b/bin/ci-bootstrap
@@ -9,9 +9,11 @@ app="${1:-hello}"
 # Run the bootstrap.yml playbook with the ansible vault password
 # passed to stdin
 echo "arnold" | docker run --rm -i \
+  -u "$(id -u)" \
   --env-file env.d/ci \
   --env K8S_AUTH_API_KEY \
   --env K8S_AUTH_HOST \
   --env OPENSHIFT_DOMAIN \
+  -v "$HOME/.kube:/home/arnold/.kube" \
   "arnold:$(tr -d '\n' < VERSION)" \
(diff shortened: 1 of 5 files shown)


bin/ci-bootstrap
```diff
@@ -9,9 +9,11 @@ app="${1:-hello}"
 # Run the bootstrap.yml playbook with the ansible vault password
 # passed to stdin
 echo "arnold" | docker run --rm -i \
+  -u "$(id -u)" \
```
lunika, line 12: I don't know if it's the best hing to do. If we want to share `~/.kube` in the container we have to make it readable. We can change the right or change the user in the container. I choose the second one, WDYT ?

jmaupetit: That's a good idea!

jmaupetit (approved):
To merge once rebased! :clap:

PR #165, opened 2018-11-26
🐛(htpasswd) generate htpassword when multiple apps are active

### LINKED ISSUE #155: create_htpasswds.yml playbook fails when multiple apps are active
## Bug Report

**Expected behavior/code**

When running the `create_htpasswds.yml` playbook, we expect creating a directory with the `htpasswd` file and each vaulted password for each active application in `group_vars/customer/{{ customer }}/{{ env_type }}/htpasswd/`.

**Actual Behavior**

When more that one application is active, only one directory is created (for the latest application in the loop I presume).

**Steps to Reproduce**

1. Activate multiple application (the `apps` var)
2. Run the `create_htpasswds.yml` playbook
3. And then the bug happens!

lunika:
## Purpose

When multiple apps are activated, only the last app treated in the
set_vars.yml task had htpasswd created. This is because the variable
containing the htpasswd path is itself in apps loop. 

Fixes #155 

## Proposal

extract the task computing the htpasswd path in a standalone task and import it everywhere we need it.

### DIFF
--- a/tasks/create_app_htpasswd.yml
+++ b/tasks/create_app_htpasswd.yml
@@ -5,6 +5,8 @@
     - secret
     - htpasswd
 
+- include_tasks: tasks/set_htpasswd_path_for_app.yml
+
 - name: Set basic auth user password files
   set_fact:
     user: "{{ item }}"
--- a/tasks/get_vault_for_app.yml
+++ b/tasks/get_vault_for_app.yml
@@ -21,10 +21,7 @@
   # Only set those variables when the vault file exists and is a regular file
   when: app_vault.stat.exists and app_vault.stat.isreg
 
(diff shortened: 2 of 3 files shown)


tasks/register_htpasswd_path_for_app.yml
```diff
@@ -0,0 +1,4 @@
+- name: Set htpasswd target file name for app
```
jmaupetit: Can you call this task set: `set_htpasswd_path_for_app.yml` instead?

lunika: Done

jmaupetit (approved):
Thx!

PR #166, opened 2018-11-27
🐛(edxapp) ignore fixture-jobs for non-trashable envs

jmaupetit:
## Purpose

When deploying `edxapp` to environments that require data persistency, we certainly don't want to reset databases or load test fixtures.

## Proposal

Hence, we ignore edxapp MySQL dump and load fixtures jobs in staging, pre-production and production.

### DIFF
--- a/apps/edxapp/templates/cms/job_05_load_fixtures.yml.j2
+++ b/apps/edxapp/templates/cms/job_05_load_fixtures.yml.j2
@@ -1,3 +1,4 @@
+{% if env_type in trashable_env_types %}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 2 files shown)


apps/edxapp/templates/cms/job_05_load_fixtures.yml.j2
```diff
@@ -1,3 +1,4 @@
+{% if env_type in trashable_env_types %}
```
sampaccoud, line 1: WDYT of adding a specific flag in configuration because some people might use this example course differently and not load it at all or always load it, or...

jmaupetit: :thinking: that's a good idea. I'll declare an issue to address this later.

sampaccoud (approved):

jmaupetit:
FTR I've declared a follow-up issue for @sampaccoud's comment (see #167).

Issue #121, opened 2018-09-11
Make blue-green deployment optional


jmaupetit:
## Purpose

Some applications don't require a blue-green deployment strategy, _e.g._ trashable services like databases for some environment, etc. We need an easy way to inactivate this and deploy them once for all.

## Proposal

Add the `has_blue_green_support` setting for each deployable app and make routes management depend on it.

PR #168, opened 2018-11-28
🔥(dc) remove deployment_stamp label for non blue-green deployments

### LINKED ISSUE #148: Remove deployment_stamp label from non blue-green deployments
## Purpose

Deploying a "non-blue-green application" with the `deployment_stamp` label in its deployment configuration leads to errors as the `deployment_stamp` changes at every deployment (OpenShift cannot patch an object that does not exist!).

## Proposal

- [ ] remove the `deployment_stamp` label of the `mailcatcher` app deployment configuration
- [ ] remove the `deployment_stamp` label of the `redis` app deployment configuration

lunika:
## Purpose

Deploying a "non-blue-green application" with the deployment_stamp label in its deployment configuration leads to errors as the deployment_stamp changes at every deployment (OpenShift cannot patch an object that does not exist!).

Fixes #148, #121 

## Proposal

- [x] remove the `deployment_stamp` label of the `mailcatcher` app deployment configuration
- [x] remove the `deployment_stamp` label of the `redis` app deployment configuration

## Update

Removing `label_stamp` is not as easy. Removing it is not enough. Every deploy is based on this stamp and the request made on OKD to find a resource is based on it. So if you just remove the stamp nothing can be deployed.

To resolve this problem and not be dependant on this stamp parameter, we have to know if an application is "stampable".
To do that we will introduce a new resource on each application, a `settings` file containing information describing the application. `databases.yml` should be merged in this new file.
To distinguish "stampable" and not "stampable" objects we will use `is_blue_green_compatible` boolean and depending on this setting we will adjust the request on OKD.

To add this settings we will modify the `apps lookup` and attach on the app object these settings

New things TODO:

- [x] create a `settings.yml` file in `apps/{app}/vars/`.
- [x] parse it in `apps` lookup and inject its content in the `app` variable.
- [x] change request on OKD to reflect this change
- [x] allow deploy of non `blue/green` application

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,7 +193,7 @@ jobs:
           name: Test the "mailcatcher" application bootstrapping
           command: |
             bin/ci-bootstrap "mailcatcher"
-            bin/ci-test-service "mailcatcher" "MailCatcher"
+            bin/ci-test-service "mailcatcher" "MailCatcher" "/" ""
 
   # Test the bootstrap playbook on the "richie" application
   # nota bene: we use a real OpenShift cluster installed in CircleCI's VM.
--- a/bin/ci-test-service
+++ b/bin/ci-test-service
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
 # usage: ci-test-service [SERVICE] [CONTENT] [PATH]
-
+set -x
 # First script argument is the application's service name. It defaults to
 # 'hello'
 service="${1:-hello}"
@@ -15,10 +15,24 @@ content="${2:-Hello OpenShift! by Arnold}"
 # Third script argument is the url path. It defaults to: "/"
 path="${3:-"/"}"
 
+# fourth script argument is the prefix used in the url. It defaults to "next"
+prefix="${4:-"next"}"
+
+# passing an empty string as prefix argument does not work, the default value (next) will
+# be assign to prefix. But we can detect how many arguments was used.
+# if exactly 4 arguments are used and prefix equals the default value then we consider that we want to use an empty string
+if [[ $# -eq 4 && "$prefix" == "next" ]]; then
+  prefix=""
+fi
+
+if [[ -n "$prefix" ]]; then
+  prefix="${prefix}."
+fi
+
 # Wait for the application to respond
 for try in $(seq 5); do
   echo "Testing ${service} application http response ($try)"
-  curl -vLk --header "Accept: text/html" "https://next.${service}.ci-eugene.${OPENSHIFT_DOMAIN}.nip.io${path}" 2> "/tmp/${service}.err" > "/tmp/${service}.out"
+  curl -vLk --header "Accept: text/html" "https://${prefix}${service}.ci-eugene.${OPENSHIFT_DOMAIN}.nip.io${path}" 2> "/tmp/${service}.err" > "/tmp/${service}.out"
   if grep "HTTP/1.1 200 OK" "/tmp/${service}.err" && grep "${content}" "/tmp/${service}.out"; then
     exit 0
   fi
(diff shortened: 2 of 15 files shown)


lunika:
The description is updated containing information of what we have to do know.

jmaupetit:
I've added a reference to #121 as this work will also fix it.

sampaccoud:
s/is_blue_green_compatible/use_blue_green ?

lunika:
@jmaupetit @sampaccoud this PR is not in WIP anymore

jmaupetit:
`use_blue_green` seems more confusing to me as we might think that an app can use blue/green or not. And this is not what we expected in the first place: this flag must stand for the blue/green compatibility.

sampaccoud:
@lunika There is another change we need to make before merging this PR. 
The `deploy` playbook should force us to specify what apps in the project we want to deploy. I always forget to specify it and the deployment fails because redis is not blue/green compatible...

After this PR is merged, the deployment will not fail anymore but it will cause a rolling-update redeploy on redis which would cause a (minor) service interruption...

lunika:
> After this PR is merged, the deployment will not fail anymore but it will cause a rolling-update redeploy on redis which would cause a (minor) service interruption...

No it will not while you don't change its deployment config, OKD is smart enough to detect no change and will not create a new deployment.

> The deploy playbook should force us to specify what apps in the project we want to deploy. I always forget to specify it and the deployment fails because redis is not blue/green compatible...

I can force to use `apps_filter` but doing this we will don't have the same default behaviour comparing to the other playbooks.

lunika:
> The deploy playbook should force us to specify what apps in the project we want to deploy. I always forget to specify it and the deployment fails because redis is not blue/green compatible...

Done

sampaccoud:
@lunika sorry I didn't answer earlier. But if pushing all apps does not trigger a redeploy unless the deployment has changed, then my comment is void and we don't need to force the user to add an `apps_filter`.

lunika:
route will be replaced because it is in `force` mode so there will be a mini down time

jmaupetit (commented):
Maybe the `forum` app should also be considered as non-blue-green-compatible? WDYT @sampaccoud ?

sampaccoud:
I don't know much about it but I would expect it should be compatible as it relies on Mongodb for persistence...

jmaupetit:
Yes it should be compatible, but do we need to deploy it as such as it will almost never get re-deployed? This adds useless complexity to its deployment IMO.

sampaccoud:
Then let's keep it blue/green but target deployments per app by default like I proposed initially?

.circleci/config.yml
```diff
@@ -286,7 +286,7 @@ jobs:
           name: Test the "marsha" application bootstrapping
           command: |
             bin/ci-bootstrap "marsha"
-            bin/ci-test-service "marsha" "Api Root" "/api/"
+            bin/ci-test-blue-green-service "marsha" "Api Root" "/api/"
```
jmaupetit: I would have preferred that you update the `bin/ci-test-service` script usage to add the blue-green option instead of creating an almost identical script.

lunika: :+1:

bin/ci-test-service
```diff
@@ -15,10 +15,26 @@ content="${2:-Hello OpenShift! by Arnold}"
 # Third script argument is the url path. It defaults to: "/"
 path="${3:-"/"}"
 
+# fourth script argument is the prefix used in the url. It defaults to "next"
+prefix="${4:-"next"}"
+
+# passing an empty string as prefix argument does not work, the default value (next) will
+# be assign to prefix. But we can detect how many arguments was used.
+# if exactly 4 arguments are used and prefix as the default value then we consider that we want to use an empty string
```
lunika: s/as/has

jmaupetit (approved):

PR #169, opened 2018-11-29
✨(bin) secure bootstrap by asking confirmation before deletion

### LINKED ISSUE #94: Secure bin/bootstrap by asking confirmation before deleting any existing project
## Feature Request

The bin/bootstrap command starts by deleting any existing project. It does not ask for confirmation so there is a risk of someone unintentionally losing information.

- [ ] Add a confirmation message before deleting the project,
- [ ] Allow forcing this confirmation by setting the "--force" or "-f" option.

lunika:
## Purpose

The bin/bootstrap command starts by deleting any existing project. It does not ask for confirmation so there is a risk of someone unintentionally losing information.

Fixes #94

## Proposal

- [x] Add a confirmation message before deleting the project,
- [x] Allow skipping this confirmation

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -1,5 +1,15 @@
 ---
 
+- hosts: local
+  gather_facts: False
+  tasks:
+    - import_tasks: tasks/set_vars.yml
+
+    - name: Ask confirmation before running bootstrap playbook to prevent deleting project accidentally
+      pause:
+        prompt: "This playbook will delete project {{ project_name }} using domain {{ domain_name }}. Press enter to continue, Ctrl+C to interrupt"
+      when: skip_verification is not defined
+
 - import_playbook: delete_project.yml
 - import_playbook: init_project.yml
 - import_playbook: build_images.yml


bootstrap.yml
```diff
@@ -1,5 +1,13 @@
 ---
 
+- hosts: local
+  gather_facts: False
+  tasks:
+    - name: Ask confirmation before running bootstrap playbook to prevent deleting project accidentally 
+      pause:
+        prompt: "This playbook will destruct existing project. Press enter to continue, Ctrl+C to interrupt"
```
sampaccoud: would be nice to display the name of the project that will be destroyed and the url of the platform ?... You might be mistaken on the environment variables you passed...

lunika: what do you mean by `url of the platform` ? The OpenShift admin console ?

lunika: If I add the information you ask, and I think this information are a good thing, I need first to compute all variables used in the project by running the task `set_vars.yml`. Doing this many thing will be prompt to user before asking him to continue or not.
Should I do this or not ?

sampaccoud: 1)I meant the domain of the OpenShift platform so you know dev from prod?
2) I would say: if you don't display this info, and with the sentence you wrote, I understand it as a generic warning but not necessarily that you HAVE FOUND an existing project with this name.

I would start the warning by: "There is already a project with the name "foo-bar"..."

lunika: We don't know if the project exists at this step. And we will don't know if it exists we just ensure that the project is absent, we don't delete it only if present.
So yes I can compute the good value after executing the `set_vars` task and then display the project and server we will work on.

bootstrap.yml
```diff
@@ -1,5 +1,15 @@
 ---
 
+- hosts: local
+  gather_facts: False
+  tasks:
+    - import_tasks: tasks/set_vars.yml
+
+    - name: Ask confirmation before running bootstrap playbook to prevent deleting project accidentally
+      pause:
+        prompt: "This playbook will destruct project {{ project_name }} using domain {{ domain_name }}. Press enter to continue, Ctrl+C to interrupt"
```
jmaupetit:
```suggestion
        prompt: "This playbook will delete project {{ project_name }} using domain {{ domain_name }}. Press enter to continue, Ctrl+C to interrupt"
```

lunika: :+1:

jmaupetit (approved):

PR #170, opened 2018-11-30
✨(vault) generate forum api key in create_vaults playbook

lunika:
## Purpose

The database API Key is not generated in the create_vaults playbook. This can be done in this playbook otherwise user must generate it afterward

## Proposal

Use lookup password to generate randomly the API Key

### DIFF
--- a/apps/forum/vars/vault/main.yml.j2
+++ b/apps/forum/vars/vault/main.yml.j2
@@ -6,10 +6,10 @@
 {% set mongodb_credentials = databases.mongodb | json_query("[?release=='" ~ forum_mongodb_version ~ "'].databases | [0][?application=='forum'].{user: user, password: password, name: name} | [0]") %}
 MONGODB_ADMIN_PASSWORD: password
 MONGODB_USER: {{ mongodb_credentials.user }}
-MONGODB_PASSWORD: {{ mongodb_credentials. password }}
+MONGODB_PASSWORD: {{ mongodb_credentials.password }}
 MONGODB_DATABASE: {{ mongodb_credentials.name }}
 
 # Forum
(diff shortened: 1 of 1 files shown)


apps/forum/vars/vault/main.yml.j2
```diff
@@ -6,10 +6,10 @@
...
 MONGODB_USER: {{ mongodb_credentials.user }}
-MONGODB_PASSWORD: {{ mongodb_credentials. password }}
+MONGODB_PASSWORD: {{ mongodb_credentials.password }}
 MONGODB_DATABASE: {{ mongodb_credentials.name }}
 
 # Forum
 # Nota bene: the forum API_KEY should match the following edxapp lms settings:
 # COMMENTS_SERVICE_KEY
-API_KEY: "thisisafakeapikey"
+API_KEY: "{{ lookup('password','/dev/null length=40') }}"
```
sampaccoud: Any idea how we could reuse the same value in the edxapp vault as they need to correspond? Cross-app secret generation...

jmaupetit: Until now I've used 48 characters-long API keys:

```suggestion
API_KEY: "{{ lookup('password','/dev/null length=48') }}"
```

jmaupetit: Maybe we should declare this improvement in an issue and implement it later?

lunika: :+1:

lunika: Maybe we can:

- decrypt the vault
- sed on the file to replace the line containing `COMMENTS_SERVICE_KEY:` with the good value
- encrypt it

But I agree with julien, in a new PR ?

jmaupetit (approved):

Issue #56, opened 2018-06-08
[CI] Publish Arnold's Docker image to Docker Hub


sampaccoud:
## Feature Request

Arnold's image takes a long time to build. It should not be a necessary step for newcomers.

### Proposal

The Continuous Integration in CircleCI should push Arnold's Docker image to Docker Hub after building it.

This has already been done on Richie here: https://github.com/openfun/richie/blob/master/.circleci/config.yml#L432

- [ ] Publish Arnold's Docker image to DockerHub when a build is triggered by a git tag
- [ ] Push an image with the same tag as the git tag AND push the same image with a "latest" tag so that it is always up-to-date.

PR #171, opened 2018-11-30
⬆️ (python) use python 3 to run arnold's playbooks

### LINKED ISSUE #78: Consider switching base image to ubuntu
## Purpose

As mentioned in #76, Arnold's docker image is a tool, so it can be ubuntu-based instead of debian-based. Switching to Ubuntu would allow to use more recent packages, particularly, python 3.6.

## Proposal

- [ ] switch to tiny `ubuntu:18.04` base image
- [ ] update apt package names and versions
- [ ] upgrade python packages versions (?)

lunika:
## Purpose

In #78 it is suggested to switch the base image used to Ubuntu. In this PR I choose python:3.6 as this image will be dedicated to tools using python.

Fixes #78

## Proposal

Upgrade to python 3.6 and adapt tools to be compatible with this version of python
For now it is not possible to upgrade to ansible 2.6 or 2.7 because the `openshift_raw` modules has been deprecated and aliased to `k8s` which is not fully compatible with `openshift_raw` (this is a bit hard for a deprecation...)

Plugins are updated to work with python 3.6, minor changes have been made.


inventory
```diff
@@ -1,3 +1,3 @@
 
-local ansible_host=127.0.0.1 ansible_connection=local
+local ansible_host=127.0.0.1 ansible_connection=local ansible_python_interpreter=/usr/local/bin/python
```
lunika: I will remove it from here. This configuration is specific to our image so it should be set on calling ansible-playbook inside container, not here.

ansible.cfg
```diff
@@ -5,5 +5,3 @@ inventory = ./inventory
 local_tmp = /tmp/ansible
 remote_tmp = /tmp/ansible
 retry_files_save_path = ~/
-callback_plugins = plugins/callback/
-stdout_callback = anstomlog
```
jmaupetit, line 9: Can you remove this callback plugin in a separated commit?

bin/ansible-playbook
```diff
@@ -8,4 +8,4 @@ source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 _set_openshift_env
 
 # Run ansible-playbook
-_docker_run ansible-playbook "$@"
+_docker_run ansible-playbook -e "ansible_python_interpreter=/usr/local/bin/python" "$@"
```
jmaupetit: Can we use an environment variable instead? I know that this configuration is not bundled with Ansible, and that we will have to handle it manually.

jmaupetit (commented):
We need to publish a docker image release using python 2.7 before merging this. Even if we do our best to test Arnold, this would be safer.

bin/ansible-playbook
jmaupetit: You can pick ideas here: https://github.com/ansible/ansible/issues/6345

ansible.cfg
lunika: :+1:

bin/ansible-playbook
lunika: `ansible_python_interpreter="/usr/bin/env python3"` in the inventory is the solution I choosed

jmaupetit:
This work is blocked until #56 has been implemented.

jmaupetit:
This PR is a major change that requires more testing. To avoid breaking our 1.0.0 release cycle, I think we need to:

1. Fix failing tests in the CI
2. Merge this PR in a `py3` branch
3. Heavily test the new python 3 implementation
4. Merge the `py3` branch in `master` once we are on the 2.0.0 roadmap.

WDYT?

jmaupetit:
After thought: this PR might not be a BC, hence this does not justify postponing it to a 2.0.0 roadmap. I'll fix and test it.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Migrate docker image to python 3.
```
```suggestion
- Migrate docker image to python 3.6
```

Dockerfile
```diff
@@ -14,15 +14,13 @@
...
 
-FROM debian:stretch
+FROM python:3.6
 
 WORKDIR /app
 
 # hadolint ignore=DL3015,DL3008
 RUN apt-get update && \
     apt-get install -y \
-    curl \
```
jmaupetit, line 24: `curl` is already installed in the `python:3.6` image, right?

jmaupetit (approved):
LGTM (after a little suggestion for the CHANGELOG).

Dockerfile
lunika: Yep, this is why I removed it.

PR #172, opened 2018-11-30
enforce database env prefix

lunika:
## Purpose

To prevent having the same prefix between different environments, we have created a mapping where all environment are defined with a corresponding prefix.

## Proposal

Create a mapping between and their prefixes and use them to generate database credentials.
Also added in this PR mongodb admin password generation.

### DIFF
--- a/apps/edxapp/vars/vault/main.yml.j2
+++ b/apps/edxapp/vars/vault/main.yml.j2
@@ -12,7 +12,7 @@ MYSQL_PASSWORD: {{ mysql_credentials.password }}
 
 # MongoDB
 {% set mongodb_credentials = databases.mongodb | json_query("[?release=='" ~ edxapp_mongodb_version ~ "'].databases | [0][?application=='edxapp'].{user: user, password: password, name: name} | [0]") %}
-MONGODB_ADMIN_PASSWORD: password
+MONGODB_ADMIN_PASSWORD: "{{ lookup('password', '/dev/null length=24') }}"
 MONGODB_USER: {{ mongodb_credentials.user }}
 MONGODB_PASSWORD: {{ mongodb_credentials.password }}
 MONGODB_DATABASE: {{ mongodb_credentials.name }}
--- a/apps/forum/vars/vault/main.yml.j2
+++ b/apps/forum/vars/vault/main.yml.j2
@@ -4,7 +4,7 @@
 
 # MongoDB
 {% set mongodb_credentials = databases.mongodb | json_query("[?release=='" ~ forum_mongodb_version ~ "'].databases | [0][?application=='forum'].{user: user, password: password, name: name} | [0]") %}
-MONGODB_ADMIN_PASSWORD: password
+MONGODB_ADMIN_PASSWORD: "{{ lookup('password', '/dev/null length=24') }}"
 MONGODB_USER: {{ mongodb_credentials.user }}
 MONGODB_PASSWORD: {{ mongodb_credentials.password }}
 MONGODB_DATABASE: {{ mongodb_credentials.name }}
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -3,6 +3,35 @@ env_type: development  # default to be overriden on command line
 customer: eugene  # default to be overriden on command line
 project_name: "{{ env_type }}-{{ customer }}"
 
+# Environments
+#
+# Each environment is defined by a name and a code. These environments can be
+# extended given your needs, but they should respect the following conventions:
+#
+# 1. the current env_type should match one of those environments name,
+# 2. an environment code should be unique among all environments.
+#
+# You should note that the environment code will be used to generate database
+# credentials with the following pattern: 
+#
+# {{ environment.code }}_{{ customer }}_{{ app.name }}
+#
+# For example, the database name for the marsha application deployed in
+# development environment for the eugene customer will be "d_eugene_marsha".
+environments:
+  - name: "ci"
+    code: "c"
+  - name: "development"
+    code: "d"
+  - name: "feature"
+    code: "f"
+  - name: "production"
+    code: "p"
+  - name: "staging"
+    code: "s"
+  - name: "preprod"
+    code: "t"
+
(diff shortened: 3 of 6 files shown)


plugins/filter/merge.py
```diff
@@ -88,7 +88,18 @@ def merge_with_database(base, database, app_name, customer, environment):
...
+    mapping_env = {
+        "ci": "c",
+        "development": "d",
+        "feature": "f",
+        "production": "p",
+        "staging": "s",
+        "preprod": "t",
+    }
+
+    env_type = mapping_env.get(environment) if environment in mapping_env else environment[0]
```
sampaccoud: You can write it like this
```python
 env_type = mapping_env.get(environment, environment[0])
```

This being said, I think this could create unexpected issues if the environment is not in your mapping but starts with the same letter as one of your environments? 

Should we make it configurable and raise an issue if the environment is not known?

lunika: Yes this can lead to an environment having the same letter. I don't know how ta make it configurable. If we want to make it configurable we have to inject in the filter I think (I will check the documentation) and I'm not really comfortable with this.

jmaupetit: Yes: it should be configurable at the Ansible level and injected while calling the filter.

```yaml
# groups_vars/all/main.yml
env_types:
  - name: "development"
    code: "d"
  - name: "staging"
    code: "s"
# [...]
```

Usage of the merge plugin should be:

```yaml
# tasks/create_app_database_vault.yml
databases: "{{ databases | default({}) | merge_with_database(item, app.name, customer, env_type, env_code) }}"
```

jmaupetit: or (better):

```yaml
# tasks/create_app_database_vault.yml
databases: "{{ databases | default({}) | merge_with_database(item, app.name, customer, env_types | json_query(name=env_type) }}"
```

The latest parameter will be a `dict` with the environment name and code.

lunika: :+1:

group_vars/all/main.yml
```diff
@@ -3,6 +3,25 @@ env_type: development  # default to be overriden on command line
 customer: eugene  # default to be overriden on command line
 project_name: "{{ env_type }}-{{ customer }}"
 
+# env_types
+# for each environment, a code is associated. This code will be used
+# to generate databases name and user with this pattern: {code}_{customer}_{app}
+# for example for the application marsha in development and the user eugene a database name will be d_eugene_marsha
+# If you need more environment you can extends this list respecting name and code keys and a unique code value
```
sampaccoud: s/extends/extend

tasks/create_app_database_vault.yml
```diff
@@ -14,6 +14,6 @@
 
 - name: Append database to databases vault
   set_fact:
-    databases: "{{ databases | default({}) | merge_with_database(item, app.name, customer, env_type) }}"
+    databases: "{{ databases | default({}) | merge_with_database(item, app.name, customer, env_types | json_query(\"[?name=='\" ~ env_type ~ \"'] | [0]\")) }}"
```
sampaccoud: I would not mind a comment explaining what the json query is selecting in plain english if its possible. I wanted to propose something but it looked like : selected the env_type from the env_types that matches the env_type :-(
A sign that the name of the `env_types` list is not optimal?

apps/edxapp/vars/vault/main.yml.j2
```diff
@@ -12,7 +12,7 @@ MYSQL_PASSWORD: {{ mysql_credentials.password }}
 
 # MongoDB
 {% set mongodb_credentials = databases.mongodb | json_query("[?release=='" ~ edxapp_mongodb_version ~ "'].databases | [0][?application=='edxapp'].{user: user, password: password, name: name} | [0]") %}
-MONGODB_ADMIN_PASSWORD: password
+MONGODB_ADMIN_PASSWORD: "{{ lookup('password','/dev/null length=24') }}"
```
jmaupetit:
```suggestion
MONGODB_ADMIN_PASSWORD: "{{ lookup('password', '/dev/null length=24') }}"
```

apps/forum/vars/vault/main.yml.j2
```diff
@@ -4,7 +4,7 @@
 
 # MongoDB
 {% set mongodb_credentials = databases.mongodb | json_query("[?release=='" ~ forum_mongodb_version ~ "'].databases | [0][?application=='forum'].{user: user, password: password, name: name} | [0]") %}
-MONGODB_ADMIN_PASSWORD: password
+MONGODB_ADMIN_PASSWORD: "{{ lookup('password','/dev/null length=24') }}"
```
jmaupetit:
```suggestion
MONGODB_ADMIN_PASSWORD: "{{ lookup('password', '/dev/null length=24') }}"
```

tasks/create_app_database_vault.yml
jmaupetit: Would you prefer `environments`? And the related comment:

> Select the environment matching the current `env_type`

group_vars/all/main.yml
jmaupetit: I would propose the following comment:

```yaml
# Environments
#
# Each environment is defined by a name and a code. These environments can be
# extended given your needs, but they should respect the following conventions:
#
# 1. the current env_type should match one of those environments name,
# 2. an environment code should be unique among all environments.
#
# You should note that the environment code will be used to generate database
# credentials with the following pattern: 
#
# {{ environment.code }}_{{ customer }}_{{ app.name }}
#
# For example, the database name for the marsha application deployed in
# development environment for the eugene customer will be "d_eugene_marsha".
```

lunika:
This is out of the scope of this PR but we could maybe rename the `validate_customer_name` task to a more generic one contain all the verification needed to run a playbook. We could add a verification that `environments` contains the current `env_type`. And verifications can be added later. WDYT ?

jmaupetit:
You can postpone this improvement, but don't forget to declare an issue for that.

jmaupetit (approved):

PR #178, opened 2018-12-03
Make volume sizes configurable

### LINKED ISSUE #173: Make volume sizes configurable
## Feature Request

Volume sizes need to be configurable.

While waiting for this feature to be available https://github.com/openfun/arnold/issues/141 we need to be able to configure the size of each volume by a setting so that customers expected to be big in production can get more place than small customers or test instances...

lunika:
## Purpose

Volumes size are not configurable, they must be configurable.

Fixes #173 

## Proposal

Define each volume size in a variable allowing to change them in the `customer`/`env_type` variable

### DIFF
--- a/group_vars/customer/eugene/development/main.yml
+++ b/group_vars/customer/eugene/development/main.yml
@@ -30,3 +30,6 @@ edxapp_theme_tag: "marvel-theme-ginkgo"
 # compile custom configMaps
 forum_host: "forum.{{ project_name}}.{{ domain_name }}"
 forum_port: 9292
+
+# increase edxapp data volume size
+edxapp_data_volume_size: 4Gi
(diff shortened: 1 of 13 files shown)


group_vars/customer/eugene/development/main.yml
```diff
@@ -30,3 +30,6 @@ edxapp_theme_tag: "marvel-theme-ginkgo"
 # compile custom configMaps
 forum_host: "forum.{{ project_name}}.{{ domain_name }}"
 forum_port: 9292
+
+# increase edxapp media volume size
+edxapp_media_volume_size: 4Gi
```
jmaupetit: I think it's the data volume that requires an increase.

jmaupetit (approved):
I am asking myself if the `edxapp_data_volume_size` shouldn't default to a higher value, else :+1:

group_vars/customer/eugene/development/main.yml
lunika: :+1:

PR #192, opened 2018-12-07
♻️ (ip) remove default cluster ip configuration

### LINKED ISSUE #181: Remove database clusters' IP addresses from variables
## Feature Request

Database clusters' IP addresses defined [in group vars](https://github.com/openfun/arnold/blob/master/group_vars/all/main.yml#L33) don't look like good defaults. 

Set these variables to null and raise an error if they are not set when deploying?

**Describe alternatives you've considered**
Maybe a better way is to define them is in each app's vault? The playbook `create_app_vaults` would include them with null as default and a comment requiring to set them.

lunika:
## Purpose

Default cluster ip configuration is not relevant because this is a configuration by customer.

Fixes #181 

## Proposal

- [x] remove default ip configuration
- [x] manage endpoint creation

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -58,11 +58,6 @@ trashable_env_types:
   - "development"
   - "feature"
 
-# Endpoints IP
-endpoint_mongodb_ip:     192.168.10.11
-endpoint_mysql_ip:       192.168.10.254
-endpoint_postgresql_ip:  192.168.10.21
-
 # Blue/Green deployment route prefixes
 blue_green_route_prefixes:
   - "previous"
@@ -114,3 +109,14 @@ http_basic_auth_users:
 aliases_port: 8999
 nginx_ports:
   - "{{ aliases_port }}"
+
+# Endpoints IP
+# These endpoints are used when you don't run your databases inside an OpenShift cluster
+# but you have dedicated external resource to manage your databases.
+# In trashable environments these endpoints will not be used because everything will
+# run inside OpenShift pods. in Production and Pre-production environments you should consider using a dedicated cluster
+# for your databases.
+# Endpoint IPs are initialized at "null" by default and you SHOULD overwrite them in the corresponding environment where you will use them.
+endpoint_mongodb_ip:     null
+endpoint_mysql_ip:       null
+endpoint_postgresql_ip:  null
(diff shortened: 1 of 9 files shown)


group_vars/all/main.yml
```diff
@@ -114,3 +109,14 @@ http_basic_auth_users:
 aliases_port: 8999
 nginx_ports:
   - "{{ aliases_port }}"
+
+# Endpoints IP
+# These endpoints are used when you don't host your databases inside OpenShift cluster
```
```suggestion
# These endpoints are used when you don't run your databases inside an OpenShift cluster
```

group_vars/all/main.yml
```diff
@@ -114,3 +109,14 @@ http_basic_auth_users:
...
+# but you have dedicated clusters to manage your databases.
```
```suggestion
# but you have dedicated external resources to manage your databases.
```

group_vars/all/main.yml
```diff
@@ -114,3 +109,14 @@ http_basic_auth_users:
...
+# Also in a trashable environment these endpoints will not be used because everything will
```
```suggestion
# In trashable environments these endpoints will not be used because everything will
```

group_vars/all/main.yml
```diff
@@ -114,3 +109,14 @@ http_basic_auth_users:
...
+# be host inside OpenShift but in Production and Preproduction you should consider using a dedicated cluster
```
```suggestion
# run inside OpenShift pods. In Production and Pre-production environments you should consider using a dedicated cluster
```

group_vars/all/main.yml
```diff
@@ -114,3 +109,14 @@ http_basic_auth_users:
...
+# for your databases.
+# They are initialized at null by default and you should overwrite them in the corresponding evenrionment you use them.
```
```suggestion
# Endpoint IPs are initialized at "null" by default and you SHOULD overwrite them in the corresponding environment where you will use them.
```

jmaupetit (approved):
:1st_place_medal: Smart move! We should fix Matsuo to support this.

group_vars/all/main.yml
lunika: :+1:

group_vars/all/main.yml
lunika: :+1:

group_vars/all/main.yml
lunika: :+1:

sampaccoud:
@lunika @jmaupetit I already updated matsuo in my 2 opened MR.

group_vars/all/main.yml
lunika: :+1:

group_vars/all/main.yml
lunika: :+1:

PR #189, opened 2018-12-05
Route deployment stamp removal

lunika:
## Purpose

route objects should not have a deployment_stamp. Having a deployment_stamp avoid massively delete an entire deployment by selecting it with its deployment_stamp because you will delete the route with this deployment_stamp and future deploy will be impossible.

Fixes #183

This PR can also partially resolve issue #188 but will also have many conflicts with #168 

## Proposal

- [x] Remove the deployment_stamp on all route objects
- [x] Update the deploy.yml playbook to change how the a deployment_stamp is retrieved
- [x] Remove the task updating the route with a new deployment_stamp and check if updating the route is still needed
- [x]  test if switch is still working

PR #193, opened 2018-12-07
🐛(routes) fix unexpected route deletion

### LINKED ISSUE #188: Routes are constantly recreated in deploy and switch playbooks
## Bug Report

When an application is deployed or routes are switched, routes are not simply patched but replaced.

This can lead to unexpected behavior. The first one I see is creating again and again a new let's encrypt certificate even if the host never change and you will reach your let's encrypt quota.

In the [ansible documentation](https://docs.ansible.com/ansible/2.5/modules/openshift_raw_module.html?highlight=openshift_raw), for the `force` parameter it is explained that if set to `true` and the resource is `present`, this resource will be replaced but the exact strategy applied is not described.

The reason I see to replace it can be found in the `delete_app` task. In this task, `Route` resources are deleted so a route (previous, current or next) can also be deleted. By replacing them on a switch, the `deployment_stamp` is updated for all routes and they cannot be deleted anymore.

IMHO this is a bug and it is critical

**Steps to Reproduce**

1. `./bin/ansible-playbook create_project.yml`
2. `./bin/init -e "apps_filter=richie"`
3. `./bin/build_images -e "apps_filter=richie"`
4. `oc annotate route richie-nginx-next bar=baz`
5. `oc annotate route richie-nginx-next foo=bar`
6. `./bin/deploy -e "apps_filter=richie`
7. at this step you don't see anymore the tag `bar=baz` on route `richie-nginx-next`
8. `./bin/switch -e "apps_filter=richie"`
9. the route `richie-nginx-current` has not the added tag anymore

**Possible Solution**

I have a working solution but more tests should should be executed, because I cannot exclude, for now, side effects.

- remove `force=true` in `deploy_patch_route.yml` task
- remove `Route` here: https://github.com/openfun/arnold/blob/master/tasks/delete_app.yml#L20 In #183 the deployment_stamp label will not be present anymore on a route object so the selector *should* never match a route so there is no reason to include `Root` in the loop

jmaupetit:
## Purpose

As precisely described in #188:

> When an application is deployed or routes are switched, routes are not simply patched but replaced.

This has major unexpected side effects.

## Proposal

The current work fixes two major issues:

1. We do not want to delete a route bound to a stack when we delete a stack with the same deployment_stamp (this is a follow-up of #189).
2. Forcing route patching during a deployment or a switch is not required and leads to erratic behavior of the OpenShift ACME application that constantly try to renew route certificate as the object changed.

Fixes #188

### DIFF
--- a/tasks/delete_app.yml
+++ b/tasks/delete_app.yml
@@ -1,4 +1,3 @@
----
 # Delete objects for the targeted deployment_stamp
 - name: Set targetted deployment stamp
   set_fact:
@@ -17,7 +16,6 @@
     - DeploymentConfig
     - Endpoints
     - Job
-    - Route
     - Service
   tags: deploy
 
--- a/tasks/deploy_patch_route.yml
+++ b/tasks/deploy_patch_route.yml
@@ -1,9 +1,7 @@
----
 # Patch routes of all services from a single app to the new deployment_stamp
 
 - name: "Patching routes with prefix {{ prefix }} for the {{ app.name }} application with deployment_stamp {{ deployment_stamp }}"
   openshift_raw:
-    force: true
     definition: "{{ lookup('template', route_template) | from_yaml }}"
     state: present
   loop: "{{ routes }}"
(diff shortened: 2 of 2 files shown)


lunika (approved):
Thank you !

jmaupetit:
:pray: I will wait for #194 to be merged before merging this PR.

sampaccoud (approved):

sampaccoud:
Small yet powerful ! 

success_kid

PR #200, opened 2018-12-11
✨(edxapp) add internationalization job

jmaupetit:
## Purpose

Default translations bundled with an edxapp release are not always up-to-date. We prefer having unvalidated translations than too many missing translations for a language.

## Proposal

The current work adds a job that uses the Transiflex client to fetch latest translations of the `edxapp` application. This leads us to add a new PVC that will store newly collected translations that will be mounted in every `edxapp` pod.


apps/edxapp/templates/cms/secret_tx.yml.j2
```diff
@@ -0,0 +1,12 @@
+{% if TX_TOKEN is defined %}
```
sampaccoud: s/tx/transifex in the name of the file?

apps/edxapp/templates/cms/secret_tx.yml.j2
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: edxapp
+    service: "cms-lms"
+  name: "{{ edxapp_tx_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  TX_TOKEN: "{{ TX_TOKEN | default('') | b64encode }}"
```
sampaccoud: s/TX_TOKEN/TRANSIFEX_TOKEN?  

The telecom engineer in me thinks Tx means "transmission" :-/

jmaupetit: `TX_TOKEN` is the expected environment variable name: https://docs.transifex.com/client/init#before-running-tx-init

I wanted to stick to that. WDYT?

sampaccoud: ok then! Maybe in the filename use transifex so it's at least explicit globally?

lunika (dismissed):

sampaccoud (approved):

jmaupetit:
To avoid making this PR too huge and mixing different purpose, I propose to merge this (once rebased), and to open a new PR dedicated to the implementation of the pre/post deployment job feature.

WDYT?

sampaccoud (approved):

PR #202, opened 2018-12-12
Create a generic pre_task to verify requirements

### LINKED ISSUE #184: Create a generic pre_task to verify requirements
## Feature Request

We created a pre_task to verify the customer name length. In fact this task should be generic to check all the requirements needed to run all our playbooks. In #172 we have a new requirement but nothing to test it.

rename the `validate_customer_name.yml` task to reflect generic validation and add the validation needed to check requirements

lunika:
## Purpose

We created a pre_task to verify the customer name length. In fact this task should be generic to check all the requirements needed to run all our playbooks. In #172 we have a new requirement but nothing to test it.

Fixes #184

## Proposal

- [x] rename validate_customer_name task in generic_validation
- [x] validate env_type is a valid environment


tasks/generic_validation.yml
```diff
@@ -2,3 +2,8 @@
   fail:
     msg: "Customer name length can not be higher than 6 characters. Current length is {{ customer | length }}."
   when: customer | length > 6
+
+- name: check if env_type is a valid environment
+  fail:
+    msg: "defined env_type {{ env_type }} is not a valid environment"
+  when: environments | json_query("[?name=='"~ env_type ~"'] | [0]") is none
```
sampaccoud: I would argue that "generic" would be a good name if it was playbook that can be run by several callers with different inputs. In this case it looks like a validation playbook with all the specific checks inside for everybody right? If yes, maybe just call it "validate_configuration"? "validate_inputs"? The use of an infinitive is in line with other playbooks: deploy, switch, etc.

tasks/generic_validation.yml
```diff
@@ -2,3 +2,8 @@
   fail:
     msg: "Customer name length can not be higher than 6 characters. Current length is {{ customer | length }}."
   when: customer | length > 6
+
+- name: check if env_type is a valid environment
```
jmaupetit:
```suggestion
- name: Check if env_type is a defined environment
```

tasks/generic_validation.yml
jmaupetit: I vote for `validate_configuration`

lunika: for now we only have configuration but this is not the purpose of this file. Its purpose is to contain generic pre_task to not have to create a new one and re-include it in all existing playbooks.

sampaccoud: My point was that the word "generic" is usually used for a shared abstraction. Here it's just a task that you play in many playbooks to "validate configuration". So it can be misleading.

lunika: there is no notion of abstraction or interface or other word with ansible.
The file is renamed.

jmaupetit (approved):

PR #203, opened 2018-12-12
Add new route for preview url

sampaccoud:
## Purpose

The preview url is used to allow instructors to review a course that is now published on the LMS. It is an accessible by intructors of the course and unlock displaying of the draft items in the course.

Since this is using a specific domain, we must create a specific route and add it to allowed hosts.

## Proposal

- [x] add a route for the preview url
- [x] add the preview host to ALLOWED_HOST in the demo project.


lunika (approved):

jmaupetit (commented):
:clap:

Issue #175, opened 2018-12-03
reuse database API KEY generated in create_vaults playbook


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

When the forum's vault is created, an API KEY is generated and this API KEY should be reused in the exdapp's vault to allow communication between those two applications. For now the only solution is to decrypt the forum's vault, copy the key, decrypt the edxapp's vault, paste the API key and then encrypt all decrypted vaults.

**Describe the solution you'd like**

In the forum's vault creation we can automate this copy/past mechanism but the generic `create_vaults` playbook is not able do that for now.
We should introduce an event system. An event is emitted at the end of a playbook allowing to create specific tasks, a more application-centric solution.

**Do you want to work on it through a Pull Request?**

sure !

lunika:
> We should introduce an event system. An event is emitted at the end of a playbook allowing to create specific tasks, a more application-centric solution.

In ansible there is a notify/handler mechanism so this is the solution I choose to implement but after spending some hours it seems that this solution is not the good one.

The idea was to create a task, looping on `apps` if a directory `handlers` exists containing a `main.yml` file. If this file exists then include it in our playbook.

There are a lot of problem with this:
- `include_tasks`, `import_tasks` and `include` can not be used because they can not import handlers but a list of tasks (yes I know it's in the name of the function but handlers are tasks). 
- You are **not** sure that your handler will be called, there are a lot of reason for that, you can fount some of them in the sporadic [documentation](https://docs.ansible.com/ansible/2.5/user_guide/playbooks_intro.html#handlers-running-operations-on-change) about handlers and you will find more with your beloved search engine.
- The handler execution is postpone at the end of the play so if something fail between the notification and the end of the play, the handler will not be executed. You can force to execute them by using `--force-handlers` option.

So I looked for another solution. A known solution is to introduce the notion of `hook` used in many modular system. A `hook` is a mounting point where you look if "modules" want to be "hook" on this point and then execute what contain the module at this point.

To implement this solution we have to create a `task` file responsible to look for a dedicated file and include it. We will have something like:

```
- include_tasks: tasks/execute_apps_hook.yml
      vars:
        hook: databases_vault_created
```

this task will be responsible to look for each defined application in `apps` if the file `apps/{{ app.name }}/hooks/databases_vault_created.yml` exists and if true include it using `include_tasks`. The tasks will be immediatly executed.

Doing this each application can have its own logic. 
[...]

PR #204, opened 2018-12-13
introduce modularity in our playbooks

lunika:
## Purpose

In #175 we want to execute a task depending on a context and specific to an app. So we need to be able to execute tasks that are "application centric".

## Proposal

Create a "hook" system. Tasks can live in the application folder and will be executed if they follow a pattern we choose. You should create a `hooks` folder and then in it files with the name of the hook you want to implement. For example for the hook `vaults_created` you should create this file for the forum application: `apps/forum/hooks/vaults_created.yml`

- [x] create tasks to execute hooks
- [x] example how to use them

### DIFF
--- a/apps/forum/hooks/vaults_created.yml
+++ b/apps/forum/hooks/vaults_created.yml
@@ -1,0 +1,5 @@
+--- 
+
+- name: manage api key
+  debug:
+    msg: "vaults created, we should inject the good api key in this hook"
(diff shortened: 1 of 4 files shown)


jmaupetit:
I like the idea, but I think we must talk a bit more of what must be achieved first. This implies dependencies between applications. IMO that should be implemented first.

lunika:
This PR is a POC and I'm open to discuss about what can be implemented

PR #207, opened 2018-12-14
Improve edxapp celery workers

jmaupetit:
## Purpose

Celery workers deployed with `edxapp` consume too much resources by default to treat too few tasks on small instances. Moreover they are not blue-green compatible since workers still running from an old stack (previous) can treat tasks related to other stacks (current or next). This last point is a major issue.

## Proposal

- [x] deploy celery workers if their queue is different from the default queue (except for the default queue worker)
- [x] suffix celery queues with the current deployment stamp to avoid tasks collision between deployed stacks

### DIFF
--- a/apps/edxapp/vars/all/main.yml
+++ b/apps/edxapp/vars/all/main.yml
@@ -60,7 +60,6 @@ edxapp_mysql_image_tag: "5.7"
 edxapp_mysql_port: 3306
 edxapp_mysql_secret_name: "edxapp-mysql-{{ edxapp_vault_checksum | default('undefined_edxapp_vault_checksum') }}"
 
-
 # -- nginx
 edxapp_nginx_image_name: "fundocker/openshift-nginx"
 edxapp_nginx_image_tag: "1.13"
@@ -69,14 +68,56 @@ edxapp_nginx_lms_port: 8071
 edxapp_nginx_htpasswd_secret_name: "edxapp-htpasswd"
 
 # -- celery/redis
-edxapp_celery_lms_high_priority_queue: "edx.lms.core.high"
-edxapp_celery_lms_default_priority_queue: "edx.lms.core.default"
-edxapp_celery_lms_low_priority_queue: "edx.lms.core.low"
-edxapp_celery_lms_high_mem_queue: "edx.lms.core.high_mem"
 
-edxapp_celery_cms_high_priority_queue: "edx.cms.core.high"
-edxapp_celery_cms_default_priority_queue: "edx.cms.core.default"
-edxapp_celery_cms_low_priority_queue: "edx.cms.core.low"
+# LMS queues
+edxapp_celery_lms_high_priority_queue: &lms_high "edx.lms.core.high"
+edxapp_celery_lms_default_priority_queue: &lms_default "edx.lms.core.default"
+edxapp_celery_lms_low_priority_queue: &lms_low "edx.lms.core.low"
+edxapp_celery_lms_high_mem_queue: &lms_high_mem "edx.lms.core.high_mem"
+
+# CMS queues
+edxapp_celery_cms_high_priority_queue: &cms_high "edx.cms.core.high"
+edxapp_celery_cms_default_priority_queue: &cms_default "edx.cms.core.default"
+edxapp_celery_cms_low_priority_queue: &cms_low "edx.cms.core.low"
+
+# LMS workers
+edxapp_celery_lms_high_priority_worker:
+  name: "lms-high"
+  queues: [*lms_high]
+  replicas: 0
+edxapp_celery_lms_default_priority_worker:
+  name: "lms-default"
+  queues:
+    - *lms_high
+    - *lms_default
+    - *lms_low
+    - *lms_high_mem
+  replicas: 1
+edxapp_celery_lms_low_priority_worker:
+  name: "lms-low"
+  queues: [*lms_low]
+  replicas: 0
+edxapp_celery_lms_high_mem_worker:
+  name: "lms-high-mem"
+  queues: [*lms_high_mem]
+  replicas: 0
+
+# CMS workers
+edxapp_celery_cms_high_priority_worker:
+  name: "cms-high"
+  queues: [*cms_high]
+  replicas: 0
+edxapp_celery_cms_default_priority_worker:
+  name: "cms-default"
+  queues:
+    - *cms_high
+    - *cms_default
+    - *cms_low
+  replicas: 1
+edxapp_celery_cms_low_priority_worker:
+  name: "cms-low"
+  queues: [*cms_low]
+  replicas: 0
 
 # -- misc
 #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Edxapp Celery workers are now configurable (number of replicas and queues to
+  consume)
+
 ### Fixed
 
 - Pods deployment wait loop is now more robust by relying on the number of pod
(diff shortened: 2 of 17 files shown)


apps/edxapp/vars/all/main.yml
```diff
@@ -69,14 +69,23 @@ edxapp_nginx_lms_port: 8071
 edxapp_nginx_htpasswd_secret_name: "edxapp-htpasswd"
 
 # -- celery/redis
-edxapp_celery_lms_high_priority_queue: "edx.lms.core.high"
-edxapp_celery_lms_default_priority_queue: "edx.lms.core.default"
-edxapp_celery_lms_low_priority_queue: "edx.lms.core.low"
-edxapp_celery_lms_high_mem_queue: "edx.lms.core.high_mem"
+#
+# To avoid deploying to much ressouces when it is not required, all queues
```
sampaccoud: s/to much ressources/several celery workers

apps/edxapp/vars/all/main.yml
```diff
@@ -69,14 +69,23 @@ edxapp_nginx_lms_port: 8071
...
+# stamp to avoid treating tasks unrelated with the currently deployed stack.
+edxapp_celery_lms_high_priority_queue: "edx.lms.core.default-{{ deployment_stamp }}"
+edxapp_celery_lms_default_priority_queue: "edx.lms.core.default-{{ deployment_stamp }}"
+edxapp_celery_lms_low_priority_queue: "edx.lms.core.default-{{ deployment_stamp }}"
+edxapp_celery_lms_high_mem_queue: "edx.lms.core.default-{{ deployment_stamp }}"
 
-edxapp_celery_cms_high_priority_queue: "edx.cms.core.high"
-edxapp_celery_cms_default_priority_queue: "edx.cms.core.default"
-edxapp_celery_cms_low_priority_queue: "edx.cms.core.low"
+edxapp_celery_cms_high_priority_queue: "edx.cms.core.default-{{ deployment_stamp }}"
```
sampaccoud: after seeing the error in CI, I think it makes more sense and solves the problem if the deployment stamp comes before the type of queue (from the most generic to the most specific):
"edx.cms.core.{{ deployment_stamp }}-default"

apps/edxapp/vars/all/main.yml
```diff
@@ -69,14 +69,23 @@ edxapp_nginx_lms_port: 8071
...
-edxapp_celery_lms_high_priority_queue: "edx.lms.core.high"
-edxapp_celery_lms_default_priority_queue: "edx.lms.core.default"
-edxapp_celery_lms_low_priority_queue: "edx.lms.core.low"
-edxapp_celery_lms_high_mem_queue: "edx.lms.core.high_mem"
+#
+# To avoid deploying to much ressouces when it is not required, all queues
+# default to the same default queue name. In this case, only the
+# DeploymentConfig corresponding to queues with a non default name will be
+# created. In other words, if you want to activate a specific queue, you will
+# need to change its name below.
```
sampaccoud: Not accurate IMO. I propose:
```
If you want to run dedicated Celery workers for a specific queue, you just need to change its name below. The template of the DeploymentConfig for this specific queue will thus be activated and specific Celery workers will be deployed for this queue.
```

sampaccoud:
Having second thoughts: actually this is not the good approach. We should keep the queues separate and differentiated.

What we want is not a single queue but a single Celery worker that consumes all the queues. It can be done with the `--queues` or `-Q` option: http://docs.celeryproject.org/en/latest/userguide/workers.html#queues

jmaupetit:
From a theoretical point-of-view, you made the point, but, at the end, the proposed solution and your alternative will lead to the same behavior: 1 consumer for all queues by default.

I need to think a bit more on how to make consumer deployment toggleable in your proposed scenario.

sampaccoud:
@jmaupetit but in your solution the queues are mixed in Redis. Maybe we need a setting defining what workers we want to and which queues they must consume. We could also define in this setting how many pods we want to spawn for each worker. WDYT?

jmaupetit:
Following your remarks, I've refactored the whole thing. And I think it's lovely now :heart_eyes:

jmaupetit:
Ready for review @sampaccoud @lunika :pray:

apps/edxapp/vars/all/main.yml
```diff
@@ -69,14 +68,56 @@ edxapp_nginx_lms_port: 8071
...
+  replicas: 0
+edxapp_celery_lms_default_priority_worker:
+  name: "lms-default"
+  queues:
+    - *lms_high
+    - *lms_default
+    - *lms_low
+    - *lms_high_mem
+  replicas: 1
+edxapp_celery_lms_low_priority_worker:
```
sampaccoud, line 96: Have you thought of wrapping this in a single `workers` variable. Then you could use a loop in the templates (I'm making the assumption that it's possible to include several objects in a k8s definition file?). Something like:
```yaml
celery_workers:
  edxapp_celery_lms_high_mem_worker:
    name: "lms-high-mem"
    queues: [*lms_high_mem]
    replicas: 0
  edxapp_celery_lms_default_priority_worker:
    name: "lms-default"
    queues:
      - *lms_high
      - *lms_default
      - *lms_low
      - *lms_high_mem
    replicas: 1
```

jmaupetit: This is interesting but too edxapp-specific IMO. If we break the one file/one object paradigm, this will require to change many things.

jmaupetit: :thinking: after-thought: I am not so sure of what I wrote. I need to think a bit more about side-effects, but having a single generic DC template for workers might be neat.

jmaupetit: Unfortunately it seems that we cannot define multiple objects in a single template (for now):

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 138, in run
    res = self._execute()
  File "/usr/local/lib/python2.7/dist-packages/ansible/executor/task_executor.py", line 511, in _execute
    self._task.post_validate(templar=templar)
  File "/usr/local/lib/python2.7/dist-packages/ansible/playbook/task.py", line 260, in post_validate
    super(Task, self).post_validate(templar)
  File "/usr/local/lib/python2.7/dist-packages/ansible/playbook/base.py", line 390, in post_validate
    value = templar.template(getattr(self, name))
  File "/usr/local/lib/python2.7/dist-packages/ansible/template/__init__.py", line 525, in template
    disable_lookups=disable_lookups,
  File "/usr/local/lib/python2.7/dist-packages/ansible/template/__init__.py", line 481, in template
    disable_lookups=disable_lookups,
  File "/usr/local/lib/python2.7/dist-packages/ansible/template/__init__.py", line 717, in do_template
    res = j2_concat(rf)
  File "<template>", line 12, in root
  File "/usr/local/lib/python2.7/dist-packages/ansible/plugins/filter/core.py", line 223, in from_yaml
    return yaml.safe_load(data)
  File "/usr/local/lib/python2.7/dist-packages/yaml/__init__.py", line 93, in safe_load
    return load(stream, SafeLoader)
  File "/usr/local/lib/python2.7/dist-packages/yaml/__init__.py", line 71, in load
    return loader.get_single_data()
  File "/usr/local/lib/python2.7/dist-packages/yaml/constructor.py", line 37, in get_single_data
    node = self.get_single_node()
  File "/usr/local/lib/python2.7/dist-packages/yaml/composer.py", line 43, in get_single_node
    event.start_mark)
ComposerError: expected a single document in the stream
  in "<unicode string>", line 6, column 1:
    apiVersion: v1
    ^
but found another document
  in "<unicode string>", line 102, column 1:
    ---
    ^
```

I'll postpone this improvement and roll back to my previous implementation.

sampaccoud (approved):

PR #211, opened 2018-12-18
🐛(htpasswd) prevent new generation if file already exists

### LINKED ISSUE #187: htpasswd are modified when deploying a fully defined project
## Bug Report

**Expected behavior/code**
on a project for which the htpasswd secrets have already been generated and committed, deploying the project to OpenShift should not modify the htpasswd files.

**Actual Behavior**
Running the "init" playbook regenerates a set of htpasswd files:
```bash
modified:   group_vars/customer/foo/production/secrets/htpasswd/edxapp/htpasswd
modified:   group_vars/customer/foo/production/secrets/htpasswd/forum/htpasswd
modified:   group_vars/customer/foo/production/secrets/htpasswd/mailcatcher/htpasswd
modified:   group_vars/customer/foo/production/secrets/htpasswd/redis/htpasswd
```

**Steps to Reproduce**
1. create a customer with all its secrets and commit all the files
2. run bin/init
3. see the modified files

**Environment**
- Arnold version:
- Platform:

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

jmaupetit:
## Purpose

On a project for which the htpasswd secrets have already been generated and committed, deploying the project to OpenShift should not modify the htpasswd files.

## Proposal

Check if file exists and do not invoke the `htpasswd` task if it does.

Fixes #187

### DIFF
--- a/tasks/create_app_htpasswd.yml
+++ b/tasks/create_app_htpasswd.yml
@@ -25,6 +25,11 @@
     - secret
     - htpasswd
 
+- name: "Check application htpasswd file status"
+  stat:
+    path: "{{ app_htpasswd_file_path }}"
+  register: app_htpasswd_file
+
 # Create a htpasswd encrypted file with random passwords for users. Password
 # files are generated plain text by default. They will be encrypted in the next
 # step. If the htpasswd file (or user password files) already exists it won't be
@@ -35,6 +40,7 @@
     name: "{{ item.user }}"
     password: "{{ lookup('password', item.path) }}"
   with_items: "{{ http_basic_auth_user_passwords | json_query('results[*].ansible_facts') }}"
+  when: not app_htpasswd_file.stat.exists
   tags:
     - secret
     - htpasswd
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Edxapp Celery workers are now configurable (number of replicas and queues to
-  consume)
+  consume) (#207)
 
 ### Fixed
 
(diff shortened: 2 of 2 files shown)


tasks/create_app_htpasswd.yml
```diff
@@ -25,6 +25,11 @@
...
     - htpasswd
 
+- name: "Check application htpasswd file status"
+  stat:
+    path: "{{ app_htpasswd_file_path }}"
+  register: app_htpasswd_file
+
 # Create a htpasswd encrypted file with random passwords for users. Password
 # files are generated plain text by default. They will be encrypted in the next
 # step. If the htpasswd file (or user password files) already exists it won't be
```
sampaccoud, line 35: If i understand well you're checking the presence of the htpasswd file only, not the presence of each password file. Note that we will need to delete the htpasswd file manually if we want to add a new user/password. I think it's not an issue but maybe remove "(or user password files)" from the comment?

sampaccoud (approved):

tasks/create_app_htpasswd.yml
jmaupetit: Password files don't need to be checked because the lookup password works as expected! Hence I think this comment is still relevant.

sampaccoud: ok then I'm misunderstanding something: why is it creating a password mismatch when the htpasswd file is being recreated? my understanding is that it is created from the passwords in the password files. So it should be deterministic?

jmaupetit: When re-running this task, the lookup password sends the password file raw (encrypted) content, and thus the `htpasswd` module thinks passwords changed and tries to update the `htpasswd` file. I think it's the most probable scenario.

PR #214, opened 2018-12-19
✨(jobs) add support for pre/post deployment jobs

jmaupetit:
## Purpose

Some jobs require to be run before deploying an application, hence we need a way to filter jobs and execute them before or after creating deployments.

## Proposal

Filtering is based on the `job_type` label, pre-deployment jobs must be labelled with `job_type: "pre"` and post-deployment jobs must be labelled with `job_type: "post"`. Jobs without `job_type` label will be considered as post-deployment jobs.

For now the only pre-deployment job is edxapp's internationalization job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Implement support for pre and post-deployment jobs
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


sampaccoud (approved):
Awesome

lunika (approved):

jmaupetit:
Once https://github.com/openfun/openedx-docker/pull/84 has been merged, we need to make a new release of openedx-docker (_e.g._ `hawthorn.1-2.3.0`) and add this tag as the default `edxapp_image_tag`, so that we ensure recent developments are not broken (due to missing dependencies).

PR #221, opened 2019-01-08
✨(volumes) add create_volumes playbook

jmaupetit:
## Purpose

When developing a new feature that requires a new volumes, one may need to create this new volume on a live project that already have been initialized.

## Proposal

We've added a create_volumes playbook that can be run during project's initialization and on a live project. We've removed the `force: true` flag on volumes creation which does not seem to be required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


sampaccoud:
If you plan to create new volumes in production, consider fixing issue https://github.com/openfun/arnold/issues/223 first?

sampaccoud (approved):

jmaupetit:
> If you plan to create new volumes in production, consider fixing issue #223 first?

I'll make a new PR for this.

PR #224, opened 2019-01-09
✨(volumes) make PV reclaim policy configurable

### LINKED ISSUE #223: Declare reclaim policies to “Retain”  for all PVC in production environment
## Bug Report

**Expected behavior/code**
When deleting a PVC (by mistake?) on a production installation, we don't want the data to be automatically deleted and lost.

**Actual Behavior**
By default Kubernetes assigns a reclaim policy of “Delete” to all PVC as explained here:
https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/#why-change-reclaim-policy-of-a-persistentvolume

**Possible Solution**
I think we should declare all our PVC with a reclaim policy of "Retain" in production.

jmaupetit:
## Purpose

By default, Kubernetes deletes a PV when a PVC is deleted by a user, leading to irreversible data loss. In some cases (_e.g._ environments), we want to keep a volume even if the corresponding PVC has been deleted.

## Proposal

- [x] Add "Retain" as the default reclaim policy for production and "Delete" for all other environments.

Fix #223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Add a new create_volumes playbook to create volumes on an existing project
+- Make PV reclaim policy configurable for all volumes
 
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


jmaupetit:
It seems that we cannot specify a PV's `persistentVolumeReclaimPolicy` from a PVC. Closing this and #223

PR #227, opened 2019-01-16
✨(edxapp) add support for extra dependencies

jmaupetit:
## Purpose

In some case, configuring edxapp for a customer/environment requires that you install extra dependencies in the Docker image that will be used (e.g. a log handler).

## Proposal

To achieve this, we decided to install those extra dependencies using our `BuildConfig` for both the LMS and CMS services.

- [x] add the `edxapp_extra_dependencies` configuration variable for `edxapp`
- [x] install extra dependencies using `pip` in the CMS & LMS BuildConfig

### DIFF
--- a/apps/edxapp/vars/all/main.yml
+++ b/apps/edxapp/vars/all/main.yml
@@ -44,6 +44,22 @@ edxapp_theme_git_private_key_secret_name: "edxapp-theme-git-sshkey-{{ edxapp_vau
 # Settings required to build your theme
 edxapp_build_settings: "fun.docker_build_production"
 
+# Customize edxapp Docker image by installing extra dependencies for a
+# customer/environment. This var is expected to be a list of dependencies
+# following the pip requirement specifier (see example below).
+#
+# NOTA BENE:
+#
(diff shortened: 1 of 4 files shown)


apps/edxapp/vars/all/main.yml
```diff
@@ -44,6 +44,16 @@ edxapp_theme_git_private_key_secret_name: "edxapp-theme-git-sshkey-{{ edxapp_vau
 # Settings required to build your theme
 edxapp_build_settings: "fun.docker_build_production"
 
+# Customize edxapp Docker image by installing extra dependencies for a
+# customer/environment. This var is expected to be a list of dependencies
+# following the pip requirement specifier (see example below).
+edxapp_extra_dependencies:
```
sampaccoud, line 56: Should we add a disclaimer for our future selves? Something like:
```
This is only intended for side packages not directly related to edxapp and specific to a customer/environment or to the way you handle operations (e.g. log handlers, monitoring agents, etc.)
```

sampaccoud (approved):

PR #229, opened 2019-01-17
✨(services) allow to define multiple services object

lunika:
## Purpose

For now, for each service in an application it is not possible to define more than one OKD `Service` object. We want to define several `Service` object like it is possible to do this with `DeploymentConfig` for example.

## Proposal

- [x] change the regex looking for `svc`files in `get_objects_for_app` task

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Allow to define multiple Service objects.
+
 ## [1.0.0-alpha.7] - 2019-01-17
 
 ### Added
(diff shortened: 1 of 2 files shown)


sampaccoud (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- allow to define multiple services object
```
sampaccoud: should be capitalized as per below's practice.

lunika: Done

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Allow to define multiple services object.
```
```suggestion
- Allow to define multiple Service objects
```

PR #246, opened 2019-02-05
⚡️(playbook) require apps_filter to switch

jmaupetit:
## Purpose

The switch playbook is supposed to run against a single application at a time due to unexpected results on non-blue-green apps or not newly deployed apps. 

## Proposal

Thus, as for the deploy playbook, we now require that the apps_filter is defined when invoking the switch playbook.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -13,12 +13,12 @@
       debug: msg="==== Starting deploy playbook ===="
       tags: deploy
 
+    - import_tasks: tasks/set_vars.yml
+
     - name: Check if apps_filter is defined
       fail:
         msg: "Deploy playbook must be used with apps_filter defined"
-      when: apps_filter is not defined
-
-    - import_tasks: tasks/set_vars.yml
+      when: apps_filter is not defined and apps | length > 1
 
     # Set the deployment stamp value for this deployment
     - name: Set deployment stamp
--- a/switch.yml
+++ b/switch.yml
@@ -18,6 +18,11 @@
 
     - import_tasks: tasks/set_vars.yml
 
+    - name: Check if apps_filter is defined
+      fail:
+        msg: "This project contains several apps, the `apps_filter` variable should be defined"
+      when: apps_filter is not defined and apps | length > 1
+
     - include_tasks: tasks/run_tasks_for_apps.yml
       vars:
         tasks:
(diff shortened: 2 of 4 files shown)


sampaccoud (approved):
Idea: check that there is something on next to switch to and fail otherwise... it seems that the current behavior is to switch even if there is nothing on `next` so we end up with nothing online...

jmaupetit:
One can think that this is a feature.

sampaccoud:
True :thinking:  but if it's a feature we should call it something else than `switch`!

switch.yml
```diff
@@ -16,6 +16,11 @@
       debug: msg="==== Starting switch playbook ===="
       tags: deploy
 
+    - name: Check if apps_filter is defined
+      fail:
+        msg: "Switch playbook must be used with apps_filter defined"
+      when: apps_filter is not defined
```
lunika: we could also check than `apps` length is higher than 1 ? If there is only 1 app defined we can skip this task.

WDYT ?

jmaupetit: Good idea! I also have to fix the deploy playbook to add this.

jmaupetit:
Ready for a new review @lunika & @sampaccoud!

lunika (approved):

sampaccoud:
I approve this change but would still like to continue the discussion on my above comment. It is a serious issue that has already caused some outage. Your argument that the switch command can be used to clear a route does not justify status-quo IMO. If we want a `clear route` feature, it should be a separate command.

sampaccoud (approved):

deploy.yml
```diff
@@ -13,12 +13,12 @@
       debug: msg="==== Starting deploy playbook ===="
       tags: deploy
 
+    - import_tasks: tasks/set_vars.yml
+
     - name: Check if apps_filter is defined
       fail:
         msg: "Deploy playbook must be used with apps_filter defined"
```
sampaccoud, line 20: Maybe change the message to reflect this new rule?
```suggestion
         msg: "This project contains several apps. Please list which apps you intend to deploy by defining the `apps_filter` option."
```

jmaupetit:
I'll handle this in a new PR. Can you please declare an issue?

sampaccoud:
Done: https://github.com/openfun/arnold/issues/248

PR #247, opened 2019-02-05
⬆️(richie) upgrade to 1.0.0-beta.1

jmaupetit:
## Purpose

We are about to deploy an official instance of Richie, thus we need to upgrade its default release in Arnold.

## Proposal

- [x] upgrade Richie to `1.0.0-beta.1`
- [x] fix jobs command (since the project's tree evolved)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Publish a Docker image to DockerHub for the master branch
 - The `apps_filter` variable definition is now required to run the `switch.yml`
   and `deploy.yml` playbooks (if multiple apps are active)
+- Upgraded Richie to its latest release (1.0.0-beta.1)
 
 ## [1.3.0] - 2019-01-31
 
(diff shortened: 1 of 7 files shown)


apps/marsha/templates/app/job_collectstatic.yml.j2
```diff
@@ -41,7 +41,11 @@ spec:
         # We point to a local registry image build for this "live" image (see
         # ImageStream and BuildConfig templates)
         image: "{{ internal_docker_registry }}/{{ project_name }}/marsha:{{ marsha_image_tag }}-live"
-        command: ["python", "manage.py", "collectstatic", "--noinput"]
```
sampaccoud: why not just:
```suggestion
command: ["python", "sandbox/manage.py", "collectstatic", "--noinput"]
```

sampaccoud (approved):

apps/marsha/templates/app/job_collectstatic.yml.j2
jmaupetit: This is easier to avoid having to modify the PYTHONPATH. But I made a mistake by modifying Masha's jobs instead of Richie's ones. :sweat_smile:

sampaccoud (approved):

lunika (approved):

PR #252, opened 2019-02-11
⬆️(core) upgrade ansible to 2.7

jmaupetit:
## Purpose

Now that we are running python 3, it's time to also update Arnold's core engine to Ansible 2.7. 

## Proposal

- [x] upgrade to Ansible 2.7.7
- [x] upgrade to openshift 0.8.4
- [x] renamed `openshift_raw` module to `k8s`
- [x] use `k8s_facts` module instead of the `k8s` lookup


jmaupetit:
Nota bene: this is a draft work for curiosity. I just want to know what fails after switching to the new `k8s` module.

sampaccoud (approved):

.circleci/config.yml
```diff
@@ -400,9 +400,55 @@ jobs:
...
+            bin/ci-test-service "richie" "Installation successful!" "/" "next"
+
+      - run:
+          name: Test "richie"'s next stack substitution (create + delete)
+          command: |
+            # Deploy the next stack a second time
+            bin/ci-ansible-playbook deploy.yml "richie"
+            # Test service re-deployed with the next route
+            bin/ci-test-service "richie" "Installation successful!" "/" "next"
+            # Test that only one stack has been deployed
```
lunika: I think this comment is not needed

bin/_config.sh
```diff
@@ -42,10 +42,13 @@ function _set_openshift_env() {
 #
 # usage: _docker_run [OPTIONS] [COMMAND]
 #
-#   OPTIONS: docker run options
-#    --env-file: file to set environment variables (_e.g._ env.d/production)
-#
-#   COMMAND: command to be run in Arnold container (_e.g._ ansible-playbook deploy.yml)
+#   OPTIONS:
+#     -e    file to set environment variables (_e.g._ env.d/production)
+#     -m    mount sources as a volume
```
lunika: You should provide an example. I have no idea how to use this option without viewing the code

bin/_config.sh
```diff
@@ -42,10 +42,13 @@ function _set_openshift_env() {
...
+#     -t    allocate a pseudo-TTY while running command
```
lunika: Same here, provide an example.

bin/_config.sh
```diff
@@ -42,10 +42,13 @@ function _set_openshift_env() {
...
+#     -v    verbose mode
```
lunika: Same here, provide an example. Should I use `-vv` or `-v=2` ?

lunika: Sorry I thought this option was the ansible verbosity.

create_redirect.yml
```diff
@@ -49,28 +49,21 @@
...
-              lookup(
-                'openshift',
-                api_version='v1',
-                namespace=project_name,
-                kind='Service',
-                resource_name='redirect-nginx'
-              )
-            ] | flatten
-          }}
+      k8s_facts:
```
lunika, line 52: "coeur avec les doigts"

delete_project.yml
```diff
@@ -15,9 +15,18 @@
 
     - import_tasks: tasks/set_vars.yml
 
+    - name: Check if project exists
```
lunika, line 20: Do you think this task is really needed ? It fails now if the project doesn't exist ?

jmaupetit: I should re-test it but yes: it failed during my first tests :cry:

lunika:
I have an issue when I want to deploy marsha. pods, jobs end in error because the env variable `DJANGO_SENTRY_DSN` does not have a real dsn (a real url I think). This is error is not visible in `CI` env because the secret has `DJANGO_SENTRY_DSN` defined. I think the line 19 of file `apps/marsha/templates/app/secret.yml.j2` should be:

```
DJANGO_SENTRY_DSN: "{{ MARSHA_VAULT.DJANGO_SENTRY_DSN | default('') | b64encode }}"
```

.circleci/config.yml
jmaupetit: Oops

delete_project.yml
jmaupetit: You are right. Maybe there was a delay before deletion and I thought nothing happens. BTW this is now fixed.

jmaupetit:
Can you declare an issue for Marsha's secret bug please?

I've addressed all of your comments in separated fixup commits. A quick review would be wonderful to make a new release :pray:

lunika:
> Can you declare an issue for Marsha's secret bug please?

Yes I can.

lunika (approved):
rebase and merge :+1:

PR #276, opened 2019-03-21
⚡️(vault) ask for vaultpass only when required

jmaupetit:
## Purpose

The vault password should only be required when creating secrets. Current configuration forced us to decrypt and load vaults variables whatever the playbook we were running.

## Proposal

- [x] remove `XXX_VAULT.*` invocation in objects that are not secrets
- [x] load application vault only when required (during secrets creation)


sampaccoud:
This is awesome :muscle:

apps/edxapp/vars/all/main.yml
```diff
@@ -15,6 +15,8 @@ edxapp_sql_dump_url: "https://gist.githubusercontent.com/jmaupetit/76fe7db4a8314
 edxapp_secret_name: "edxapp-{{ edxapp_vault_checksum | default('undefined_edxapp_vault_checksum') }}"
 
 # i18n
+# Should we download and compile latest translations?
+edxapp_i18n_should_update: false
```
sampaccoud: Maybe start with `should` then the verb?
```suggestion
edxapp_should_update_18n: false
```

apps/marsha/templates/app/dc.yml.j2
```diff
@@ -61,19 +61,15 @@ spec:
...
           persistentVolumeClaim:
             claimName: marsha-pvc-media
         - name: marsha-v-static
           persistentVolumeClaim:
             claimName: marsha-pvc-static
-{% if MARSHA_VAULT.DJANGO_CLOUDFRONT_PRIVATE_KEY is defined %}
         - name: marsha-cloudfront-private-key-secret
           secret:
             secretName: "{{ marsha_cloudfront_private_key_secret_name }}"
-{% endif %}
```
sampaccoud: just curious: do you remove it because in our case we always use it or because it works even if it is not set?

jmaupetit: It is required, so I expect the secrets creation to fail if the vault entry does not exists.

apps/edxapp/vars/all/main.yml
jmaupetit: Ah ah. I knew this was a bad name (talked about it with @lunika), but I kind like namespacing vars as `{{ application.name}}_{{ scope }}_{{ purpose }}`. But I can ignore this implicit rule and update this variable name. WDYT?

apps/marsha/templates/app/dc.yml.j2
sampaccoud: actually in Marsha it's not required if you don't activate URL signing in CloudFront. I'm not sure though whether we fully planned this possibility in Arnold.

apps/edxapp/vars/all/main.yml
sampaccoud: I would say namespacing with application name is fine. Scope may be too much since we have only a few settings?

jmaupetit: Ok then, I'll do this.

apps/marsha/templates/app/dc.yml.j2
jmaupetit: @lunika did. Anyway, remove request signature is not something we would recommend right?

lunika: No and by default you cannot change this setting in our terraform so cloudfront will be configured to restrict viewer access. You have to change this setting in the cloudwatch interface.

sampaccoud: so it's a half baked thing in Marsha because I put some Django settings to deactivate it :disappointed:. The flag in cloudfront would be easy to add. Should we create an issue for that?

jmaupetit: Yes! Both here and in Marha's repository.

sampaccoud: ok let me do it.

apps/marsha/templates/app/secret_cloudfront_private_key.yml.j2
```diff
@@ -1,4 +1,3 @@
-{% if MARSHA_VAULT.DJANGO_CLOUDFRONT_PRIVATE_KEY is defined %}
```
jmaupetit, line 1: Maybe I should rollback this to make `DJANGO_CLOUDFRONT_PRIVATE_KEY` optional in Marsha's vault? Ping @lunika

apps/marsha/templates/app/dc.yml.j2
jmaupetit: I think it's only required for Marsha, I've rolled back this feature in Arnold.

lunika (approved):

PR #283, opened 2019-03-26
🔧(pyup) switch to weekly updates

jmaupetit:
## Purpose

By default, pyup submits a new MR for every new dependency update. We prefer scheduling updates with a weekly frequency.

## Proposal

- [x] add `.pyup.yml` configuration file


sampaccoud (approved):
:+1:

Issue #308, opened 2019-05-03
Add tray version support


jmaupetit:
## Purpose

While fetching an application tray, we need to support multiple versions of the tray.

## Proposal

- [ ] update the apps tree to something like: 

`apps/{{ tray.name }}/{{ tray.version }}/templates/...`

- [ ] add versions to active apps, _e.g._:

```yaml
apps:
  - name: richie
    version: 1.0.0
```
- [ ] collect apps given this version

Optional:

- [ ] add default version support

jmaupetit:
Not sure this is relevant in fact. Will we deploy two different tray releases in the same project ?

PR #309, opened 2019-05-03
Add community applications (trays) fetch support

jmaupetit:
## Purpose

Now that we are able to work with tray-packaged apps, the fun part begin! We now need to be able to download, configure and add community apps that are not part of official apps.

## Proposal

- [x] add `community_apps` target directory
- [x] add `fetch_dependencies.yml` playbook and arnold `install` command


group_vars/all/apps.yml
```diff
@@ -0,0 +1,22 @@
+#
```
jmaupetit: I should annotate/comment this file.

group_vars/all/apps.yml
```diff
@@ -0,0 +1,22 @@
...
+app_tray_finder: "arnold.yml"
+
+# Paths to look for deployable applications
+apps_paths:
+  - "apps/"
+  - "community_apps/"
+
+apps_repository:
+  - name: funmooc
```
jmaupetit: Should comment this as a template

group_vars/all/apps.yml
```diff
@@ -0,0 +1,22 @@
+#
+app_tray_finder: "arnold.yml"
+
+# Paths to look for deployable applications
+apps_paths:
+  - "apps/"
+  - "community_apps/"
+
+apps_repository:
```
jmaupetit:
```suggestion
apps_registry:
```

group_vars/all/apps.yml
```diff
@@ -0,0 +1,22 @@
...
+  - name: funmooc
+    git:
+      # Should be https or ssh (requires more configuration) git repository url
+      repository: https://github.com/openfun/fun-mooc.git
+      # Can be a tag, a branch or a commit sha1
+      version: add-arnold-package
```
jmaupetit: This should be guessed from the application version listed in `apps` (when #308 will be implemented).

group_vars/all/apps.yml
```diff
@@ -0,0 +1,22 @@
...
+
+core_apps_paths:
+  - "core_apps/"
+
+community_apps_dir: "community_apps"
```
jmaupetit: Will be removed once all apps moved from Arnold to their own repository.

tasks/fetch_app.yml
```diff
@@ -0,0 +1,48 @@
...
+- name: Remove old tray
+  file:
+    path: "{{ tray_dest_path.stat.path }}"
+    state: absent
+  when: tray_dest_path.stat.exists and tray_dest_path.stat.isdir
+
+- name: Install fetched tray
+  copy:
+    src: "{{ build_dir.path }}/{{ tray_finder.tray.path }}/"
+    dest: "{{ community_apps_dir }}/{{ tray.metadata.name }}"
```
lunika: You can maybe also removed the temporary clone ?

group_vars/all/apps.yml
lunika: I'm not sure this file should be versionned. Maybe I'm wrong but I understand that I will have to edit it if I want to add a third party application. If I modify it, it will appear in the list of modified files in my git status and I don;t think we want this.

fetch_apps.yml
```diff
@@ -0,0 +1,15 @@
+---
+- hosts: local
+  gather_facts: False
+
+  tasks:
+    - name: Display playbook name
+      debug: msg="==== Starting fetch_apps playbook ===="
+
+    - import_tasks: tasks/set_vars.yml
```
lunika: Is it really needed ?

jmaupetit: I have to test it but probably not.

tasks/fetch_app.yml
jmaupetit: Not sure, that is the OS responsibility to delete temporary files (in `/tmp`) not our.

group_vars/all/apps.yml
jmaupetit: You can safely override default variables for a customer/env_type with your own apps repository.

bin/arnold
```diff
@@ -194,6 +200,7 @@ COMMANDS:
   create_db_vault   create or update a vault file with databases credentials
   htpasswd          initialize a namespace's htpassword files
   info              display current configuration
+  install           fetch dependencies (deployable community apps)
```
lunika: `install` is not really intuitive. I expect something else then fetching dependencies.

Could be `dependencies` ? `dependency` ?

lunika (commented):
We will have to manually maintain a registry in this repo ?

bin/arnold
jmaupetit: Would `install_trays` be more explicit?

jmaupetit:
> We will have to manually maintain a registry in this repo ?

Yup. Until we come up with a better idea.

bin/arnold
lunika: yes

fetch_dependencies.yml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Re-check application registry
+      stat:
+        path: "{{ apps_registry_file_name }}"
+      register: registry_file
+
+    - name: Load registry
+      include_vars:
+        file: "{{ apps_registry_file_name }}"
+        name: registry
+      when: registry_file.stat.exists
```
lunika, line 37: You should also abort if the registry file is not created.

fetch_dependencies.yml
```diff
@@ -0,0 +1,63 @@
...
+        - name: Abort if dependencies file does not exist
+          debug:
+            msg: "No dependencies file found"
+        - name: Exit playbook
+          meta: end_play
+      when: not dependencies_file.stat.exists
+
+    - name: Load dependencies
+      include_vars:
+        file: "dependencies.yml"
```
lunika: can we variabilize this file name ?

registry.yml
```diff
@@ -0,0 +1,20 @@
+# Community apps registry
```
lunika, line 1: Did you plan to check if this file is correctly formatted ? A doc somewhere with properties we can use ? `subpath` is not always used as I see.

tasks/fetch_app.yml
```diff
@@ -0,0 +1,102 @@
...
+    verbosity: 2
+
+- name: Check already installed tray
+  stat:
+    path: "{{ tray_dest_path }}"
+  register: tray_dest
+
+- name: Confirm tray update
+  pause:
+    prompt: "{{ tray.metadata.name }} tray already exists. Do you want to update it? Press enter to continue, Ctrl+C to interrupt"
```
lunika: `Ctrl+c` will interrupt all the command or just this app installation ? If I install 3 apps, the first is already installed and I don't want to update it, is it possible to skip it and continue ?

lunika (approved):

lunika (changes requested):
Sorry, I don't want to block the PR but I can't undo my previous approval and it was a mistake.

tasks/fetch_app.yml
jmaupetit: Unfortunately it will break the loop, so interrupt all the command.

fetch_dependencies.yml
jmaupetit: Sure we can, but it's more like a convention.

registry.yml
jmaupetit: Good points. I should definitely write a documentation for this. What do you have in mind for its validation? A schema?

tasks/fetch_app.yml
lunika: I think we should not. Is there another way to do it in multiple step ? Asking a value and depending on this value skip remaining tasks ?

registry.yml
lunika: Yes like a schema and validated in the CI

tasks/fetch_app.yml
jmaupetit: I've implemented this behavior. Thanks for the suggestion. :muscle:

registry.yml
jmaupetit: I've added a JSONSchema specification for the registry file format, and it is now automatically checked by the CI.

jmaupetit:
@lunika I've addressed all your comments. Ready for a last review? :pray:

lunika (approved):

PR #310, opened 2019-05-06
✨(plugins) allow application template override

jmaupetit:
## Purpose

We often need to redefine a single template file for a tray, _i.e._ to customize a secret or a deployment for a customer/environment.

## Proposal

We made this possible using the `merge_with_app` filter: considering that we want to override only Richie's application secret, one will need to declare a new secret template and override the secret path in the customer/environment variables file (_e.g._ `group_vars/customer/foo/staging/main.yml`):

```yaml
apps:
  - name: richie
    services:
      - name: app
        templates:
          - group_vars/customer/fun/apps/richie/templates/richie/secret.yml.j2
```

This will override the default secret template (refered as `apps/richie/templates/richie/secret.yml.j2`).

### DIFF
--- a/filter_plugins/merge.py
+++ b/filter_plugins/merge.py
@@ -3,6 +3,7 @@ Merge Jinja filters
 """
 
 from copy import deepcopy
+from pathlib import Path
 
 from ansible.errors import AnsibleFilterError
 from ansible.utils.encrypt import random_password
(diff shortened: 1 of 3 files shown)


filter_plugins/merge.py
```diff
@@ -51,8 +52,26 @@ def merge_with_app(base, new):
...
+                # as they both share the same file name, e.g. foo.yml, the new
+                # one will be used and the old one ignored.
+                base_service[k] = sorted(
+                    list(
+                        set(
+                            new_service[k]
+                            + [
+                                f
+                                for f in base_service[k]
+                                if Path(f).name not in new_service_files
```
sampaccoud: not sure how there could be duplicates since you are checking that the file name if not in new_service[k].

jmaupetit: True. There cannot be duplicates anymore.

sampaccoud (approved):

PR #312, opened 2019-05-07
🐛(playbook) fix display name for new projects

jmaupetit:
## Purpose

When creating a new project using the `create_project.yml` playbook, the project display name field is empty.

## Proposal

- [x] fix the expected "display name" field name
- [x] make project definition more explicit

FTR the Project object schema reference is available at:
https://docs.openshift.com/online/rest_api/apis-project.openshift.io/v1.Project.html#object-schema

### DIFF
--- a/create_project.yml
+++ b/create_project.yml
@@ -17,13 +17,14 @@
 
     - name: Make sure the project exists in OpenShift and is up-to-date
       k8s:
-        api_version: v1
-        kind: Project
-        name: "{{ project_name }}"
-        resource_definition:
-          description: "{{ project_description|default(project_display_name)|default(project_name) }}"
-          display_name: "{{ project_display_name|default(project_name) }}"
         state: present
+        definition:
+          apiVersion: project.openshift.io/v1
+          kind: Project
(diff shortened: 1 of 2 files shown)


create_project.yml
```diff
@@ -17,13 +17,14 @@
 
     - name: Make sure the project exists in OpenShift and is up-to-date
       k8s:
-        api_version: v1
-        kind: Project
-        name: "{{ project_name }}"
-        resource_definition:
-          description: "{{ project_description|default(project_display_name)|default(project_name) }}"
-          display_name: "{{ project_display_name|default(project_name) }}"
```
sampaccoud, line 25: There seem to be a change in the format as well that is not described in the commit message?

jmaupetit: `definition` is an alias for `resource_definition`. I preferred moving all fields for this object as `definition` field children instead of mixing `definition` and `k8s` Ansible module arguments. This change is stated in the commit message by:

> changed the object definition to make it more explicit.

sampaccoud (approved):

PR #333, opened 2019-06-12
⚡️(apps) increase healthchecks periodicity

jmaupetit:
## Purpose

When a database service is not available, the frequency of our pods healthcheck is so high that our platform suffers then from a heavy load.

## Proposal

- [x] increase liveness/readiness probes `periodSeconds`

### DIFF
--- a/apps/edxapp/templates/services/cms/_dc_base.yml.j2
+++ b/apps/edxapp/templates/services/cms/_dc_base.yml.j2
@@ -93,7 +93,7 @@ spec:
               - "-c"
               - "python manage.py {{ service_variant }} celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
           initialDelaySeconds: 120
-          periodSeconds: 5
+          periodSeconds: 30
         readinessProbe:
           exec:
             command:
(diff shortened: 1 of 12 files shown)


apps/edxapp/templates/services/cms/_dc_base.yml.j2
```diff
@@ -93,15 +93,15 @@ spec:
...
+          periodSeconds: 30
         readinessProbe:
           exec:
             command:
               - "/bin/bash"
               - "-c"
               - "python manage.py {{ service_variant }} celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
           initialDelaySeconds: 60
-          periodSeconds: 3
+          periodSeconds: 10
```
lunika, line 104: this can maybe increase a little bit more as we know this check consume a lot ?

jmaupetit: This is a readiness probe, not liveness.

lunika: :+1:

lunika (approved):

PR #337, opened 2019-06-20
✨(playbook) ensure secrets exist before deploying

jmaupetit:
## Purpose

When updating a vault content, we often forget to regenerate our secrets before re-deploying an application. In this case, the deployment fails and we need to clean up our newly created orphan stack.

To prevent this failure, @lunika suggested that we check secrets before deploying.

## Proposal

- [x] Add a new `check_app_secrets` task set

### DIFF
--- a/tasks/check_app_secrets.yml
+++ b/tasks/check_app_secrets.yml
@@ -1,0 +1,39 @@
+# Check expected app secrets exists. This is useful to avoid deploying an
+# application for which secrets have not been created first.
+
+- name: Does current app has secrets?
+  set_fact:
+    app_has_secrets: "{{
+      app |
+      json_query('services[*].templates[]') |
+      map('regex_search', '.*/secret.*\\.yml\\.j2$') |
+      select('string') | list | length > 0
+    }}"
+  tags: secret
+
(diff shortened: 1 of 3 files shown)


tasks/check_app_secrets.yml
```diff
@@ -0,0 +1,39 @@
...
+    api_version: "v1"
+    namespace: "{{ project_name }}"
+    kind: Secret
+    label_selectors:
+      - app={{ app.name }}
+  register: existing_secrets
+  when: app_has_secrets
+  tags: secret
+
+- name: "Filter secrets matching expected checksum {{ app_vault_checksum }}"
```
jmaupetit: Not sure that `app_vault_checksum` is always matching the current application if `apps_filter` contains multiple apps. I'll improve this.

jmaupetit: Fixed.

sampaccoud:
Why do I feel stigmatized by this PR? Good move @lunika ^^

jmaupetit:
>Why do I feel stigmatized by this PR?

We are all on the same boat :wink:

lunika (approved):

PR #338, opened 2019-06-20
✨(playbook) add delete_previous playbook

jmaupetit:
## Purpose

Being able to remove an old previous stack is useful to lower k8s cluster load. 

## Proposal

This new playbook should be used with caution only for stacks that will never be switched back.


delete_previous.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Display playbook name
+      debug: msg="==== Starting delete_previous playbook ===="
+      tags: deploy
+
+    - import_tasks: tasks/set_vars.yml
+
+    - name: Check if apps_filter is defined
+      fail:
+        msg: "This project contains several apps, the `apps_filter` variable should be defined"
+      when: apps_filter is not defined and apps | length > 1
```
jmaupetit: This task is used in 3 different playbooks, should we extract it in a new task set and import it? Ping @lunika

lunika: I think you can, yes.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Optimize `edxapp` workers liveness probe frequency and replication to lower
   base cluster load
 - Ensure expected secrets exists before deploying an application
+- Add `delete_previous.yml` playbook to delete previous stacks (should be used
+  with caution)
```
jmaupetit, line 15: Wrong section.

lunika (approved):

PR #339, opened 2019-06-20
✨(playbook) add blue-green rollback

### LINKED ISSUE #134: Add a switch-back playbook
## Purpose

In some cases, being able to restore the previous state to the current one would be a lovely feature.

## Proposal

Following the `switch.yml` playbook, we need to:

- switch previous to current
- switch current to next
- delete next (if any)

jmaupetit:
## Purpose

Using a blue-green deployment strategy, we need to be able to easily rollback a previous stack as the current one.

## Proposal

- [x] add `rollback.yml` playbook

Fix #134

### DIFF
--- a/rollback.yml
+++ b/rollback.yml
@@ -1,0 +1,28 @@
+---
+# This playbook rolls back all routes:
+# 1. point the "current" route to the stack that was pointed by the "previous" route,
+# 2. point the "next" route to the stack that was pointed by the "current" route,
+
+- hosts: local
+  gather_facts: False
+
+  pre_tasks:
+    - import_tasks: tasks/check_configuration.yml
+    - import_tasks: tasks/check_apps_filter.yml
+
+  tasks:
+    - name: Display playbook name
+      debug: msg="==== Starting rollback playbook ===="
+      tags: rollback
+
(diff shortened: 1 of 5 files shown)


rollback.yml
```diff
@@ -0,0 +1,27 @@
...
+  gather_facts: False
+
+  pre_tasks:
+    - import_tasks: tasks/check_configuration.yml
+    - import_tasks: tasks/check_apps_filter.yml
+
+  tasks:
+    - name: Display playbook name
+      debug: msg="==== Starting rollback playbook ===="
+      tags: deploy
```
lunika:
```suggestion
      tags: rollback
```

rollback.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    - import_tasks: tasks/set_vars.yml
```
lunika, line 18: You should also add the `rollback` tag or it will skip it if you use the tag.

lunika (commented):
Can you add a helper in the `bin` directory to ease the use of this playbook ?
Thanks.

jmaupetit:
> Can you add a helper in the bin directory to ease the use of this playbook ?

Done.

lunika (approved):

PR #348, opened 2019-07-15
⚡️(edxapp) decrease celery worker probes load

jmaupetit:
## Purpose

We still have heavy load issues in production due to our celery worker probes.

## Proposal

Using celery commands wrapped in a Django management command leads to an heavy load that is not suitable to be run for pod probes. We believe that this heavy load is mostly due to the huge edx Django stack that should perform some checking while loading.

- [x] use configured/native `celery` commands
- [x] performance live checks every 15 minutes

### DIFF
--- a/apps/edxapp/templates/services/cms/_dc_base.yml.j2
+++ b/apps/edxapp/templates/services/cms/_dc_base.yml.j2
@@ -91,15 +91,15 @@ spec:
             command:
               - "/bin/bash"
               - "-c"
-              - "python manage.py {{ service_variant }} celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
+              - "celery -b redis://redis:{{ redis_app_port }} inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
           initialDelaySeconds: 120
-          periodSeconds: 400
+          periodSeconds: 600
         readinessProbe:
           exec:
             command:
               - "/bin/bash"
               - "-c"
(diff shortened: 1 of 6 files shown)


apps/edxapp/templates/services/cms/_dc_base.yml.j2
```diff
@@ -91,15 +91,15 @@ spec:
             command:
               - "/bin/bash"
               - "-c"
-              - "python manage.py {{ service_variant }} celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
+              - "celery -b redis://redis:{{ redis_app_port }} inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)"
           initialDelaySeconds: 120
-          periodSeconds: 400
+          periodSeconds: 900
```
sampaccoud: Maybe you don't need to increase it anymore now that you optimized it?

jmaupetit: Yup, but I think that: i. we should have a period that corresponds to a minute-based granularity (_e.g._ 10 minutes), and, ii. we agreed that given the number of replicates and pods affinity, something like `x>=10` minutes should be ok to restart a worker that handles asynchronous jobs. Can we agree on `600` seconds?

sampaccoud: ok

sampaccoud (approved):

PR #353, opened 2019-07-19
✨(marsha) allow defining resources requests for all pods

sampaccoud:
## Purpose

We want to improve pod scheduling on our OpenShift clusters. 

## Proposal

By setting resources requests without any limit, our pods are in "Burstable" mode: we can define the minimum resources it requires upon scheduling, while still allowing it to burst.

### DIFF
--- a/apps/marsha/vars/all/main.yml
+++ b/apps/marsha/vars/all/main.yml
@@ -38,3 +38,21 @@ marsha_activate_http_basic_auth: false
 # -- volumes
 marsha_media_volume_size: 2Gi
 marsha_static_volume_size: 2Gi
+
+# -- resources
+marsha_app_resources:
+  requests:
+    cpu: 10m
+    memory: 250Mi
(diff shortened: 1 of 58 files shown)


apps/marsha/vars/all/main.yml
```diff
@@ -37,3 +37,11 @@ marsha_should_sign_requests: true
 # -- volumes
 marsha_media_volume_size: 2Gi
 marsha_static_volume_size: 2Gi
+
+# -- resources requests
+marsha_resources_requests_app_cpu: 10m
```
jmaupetit: What about the following naming convention?

```
{{ app.name }}_{{ service}}_resources_cpu: 10m
```

sampaccoud: you're right the service should be first. As for `requests`, we need to keep it because resources limits exist alongside resources requests.

jmaupetit: Ok then.

madmatah:
Replaced by #644

PR #359, opened 2019-08-19
⚡️(redis) improve healthcheck reliability

jmaupetit:
## Purpose

When for some reason the `/data` mounted volume is lost, `redis` is not able to dump the current database on disk and it is no longer responding to requests. This is a SPOF as other apps crashes subsequently (_e.g._ `edxapp`).

## Proposal

We now check that we are able to write to `/data` to ensure that the volume is properly mounted with write permissions.

### DIFF
--- a/apps/redis/templates/services/app/dc.yml.j2
+++ b/apps/redis/templates/services/app/dc.yml.j2
@@ -30,7 +30,7 @@ spec:
               command:
                 - "/bin/bash"
                 - "-c"
-                - "redis-cli ping"
+                - "'redis-cli ping && touch /data/healthcheck'"
             initialDelaySeconds: 120
             periodSeconds: 30
           readinessProbe:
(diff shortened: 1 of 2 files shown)


apps/redis/templates/services/app/dc.yml.j2
```diff
@@ -30,15 +30,15 @@ spec:
               command:
                 - "/bin/bash"
                 - "-c"
-                - "redis-cli ping"
+                - "'redis-cli ping && touch /data/healthcheck'"
```
sampaccoud, line 33: Should we delete it after to make sure it is created each time?

jmaupetit: I don't think so because you cannot touch something if you don't have the permission to touch it :sweat_smile: (I mean write permission, because `touch` updates touched file metadata). In other words, if the file already exists in a read-only volume or a volume that is no longer mounted, you will get an error; and this is what we expect.

sampaccoud (approved):

PR #366, opened 2019-09-11
✨(marsha) allow to configure the command running marsha contrainer

lunika:
## Purpose

Gunicorn in docker need few modifications to run well an application
(see https://pythonspeed.com/articles/gunicorn-in-docker/). For marsha
we want to apply some of these rules, the easiest and flexible way to do
this is to override the command running in the container and set it in a
variable.

## Proposal
- [x] create a variable to set the gunicorn command
- [x] use the command property in the DC template for the marsha container

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Upgrade OpenShift to OpenShift `0.9.2`
+- Set in a variable the command running marsha container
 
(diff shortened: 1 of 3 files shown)


jmaupetit (changes requested):
As discussed IRL, I would prefer that we modify [Marsha's Gunicorn configuration file](https://github.com/openfun/marsha/blob/master/docker/files/usr/local/etc/gunicorn/marsha.py) instead and not mixing command line override an flat file settings. By doing so, in a near future, we will be able to add a ConfigMap for this file in Arnold.

PR #371, opened 2019-09-25
switch learning locker to blue/green stategy

### LINKED ISSUE #369: switch learning locker to blue/green stategy
## Feature Request

Learning locker application does not use the blue/green strategy. We think that we can use this strategy, learning locker should be compatible with it

- remove the setting `is_blue_green_compatible`
- create a job to run MongoDB migrations (http://docs.learninglocker.net/guides-cli/#migratemongo)

Another thing to add but out of the scope of this issue:

- configure replicas number

lunika:
## Purpose

Blue/green deployment for learning locker is deactivated. We want to activate it. All templates should be rework and we also want to introduce a new job responsible to run mongo migrations after a deployment.
There is also a problem with the mongo endpoint. Each deploy should create a new endpoint object and like others objects suffixed with the deployment stamp. To connect to mongo, learning locker uses an environment variable named `MONGODB_PATH` which is the full Mongo [connection string](https://docs.mongodb.com/manual/reference/connection-string/). The mongo domain must be part of this connection string and it changes for each deployment. Setting it in the secrets is not possible anymore and further more I need information stored in the application vault in the deploy task. To handle this, we decided to set in the secret mongo's credentials and then craft the `MONGODB_PATH` variable in the DC template using variable expansion explained in this documentation: https://github.com/kubernetes/kubernetes/blob/aec7b416eb82c5a4c43ac786f02539a5669a1897/docs/design/expansion.md

## Proposal

- [x] refactor learning locker templates to made them blue/green compatible
- [x] allow to configure number of replicas for each applications (api, ui, worker and xapi)
- [x] create a job to run mongo migrations

Fixes #369


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
```
jmaupetit, line 12: In your commit message: `s/runa/run a/`

apps/learninglocker/templates/services/app/_dc_base.yml.j2
```diff
@@ -11,7 +11,7 @@ metadata:
   name: "{{ dc_name }}-{{ deployment_stamp }}"
```
jmaupetit, line 11: In your commit title: `s/each DCs/each DC/`

tasks/create_app_config.yml
```diff
@@ -14,6 +14,11 @@
   set_fact:
     app_configmaps: "{{ app | json_query('services[?configs].{name: name, configs: configs}') }}"
 
+- include_tasks: tasks/run_tasks_for_app.yml
+  vars:
+    tasks:
+      - tasks/get_vault_for_app
```
jmaupetit: This cannot work, as we (or more precisely the deployment bot) does not have access to the vault during the deployment.

jmaupetit (commented):
I would suggest to inject environment variables from the secret (MongoDB credentials) and use them in the ConfigMap's  `MONGODB_PATH` environment variable that will be lazily interpreted, _e.g._:

```yaml
MONGODB_PATH: mongodb://${MONGODB_USER}:${MONGODB_PASSWORD}@{{ learninglocker_mongodb_host }}-{{ deployment_stamp }}:{{ learninglocker_mongodb_port }}/${MONGODB_DATABASE}
```

apps/learninglocker/templates/services/app/_dc_base.yml.j2
```diff
@@ -34,6 +36,9 @@ spec:
                 name: "{{ learninglocker_secret_name }}"
             - configMapRef:
                 name: "learninglocker-app-dotenv-{{ deployment_stamp }}"
+          env:
+            - name: MONGODB_PATH
+              value: "mongodb://$(MONGODB_USER):$(MONGODB_PASSWORD)@{{ learninglocker_mongodb_host }}-{{ deployment_stamp }}:{{ learninglocker_mongodb_port }}/$(MONGODB_DATABASE)"
```
jmaupetit, line 41: :1st_place_medal:

jmaupetit (approved):

PR #376, opened 2019-10-04
Fix redis healthchecks

jmaupetit:
## Purpose

The health checks command fails, so we are unable to deploy `redis` as it is never ready.

## Proposal

- [x] remove extra quotes from `redis` pod health checks


sampaccoud (approved):

sampaccoud:
worth a :beers:

PR #382, opened 2019-10-09
🐛(secrets) fix private docker registry task when checking if file exists

sampaccoud:
## Purpose

The condition was reversed when checking if a secrets file exists with credentials to private Docker registries.

## Proposal

I modified the condition to proceed when a secrets file exists and stop otherwise.


jmaupetit (approved):
Sorry daddy :sweat:

PR #385, opened 2019-10-15
🐛(edxapp) fix jobs image name when using custom theme

rmoch:
## Purpose

When using a custom theme, `edxapp` jobs containers are supposed to run using a non-existing image since the theme version is included in the image name reference.

## Proposal

- [x] use the `image_name` macro in `edxapp` jobs

### DIFF
--- a/apps/edxapp/templates/services/cms/job_01_internationalization.yml.j2
+++ b/apps/edxapp/templates/services/cms/job_01_internationalization.yml.j2
@@ -1,3 +1,4 @@
+{%- from "apps/edxapp/templates/services/cms/macros/image.yml.j2" import image_name -%}
 {% if edxapp_should_update_i18n %}
 apiVersion: batch/v1
 kind: Job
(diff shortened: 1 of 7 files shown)


apps/edxapp/templates/services/cms/_bc_base.yml.j2
```diff
@@ -18,7 +18,7 @@ metadata:
   # References:
   #
   # 1. https://docs.okd.io/latest/dev_guide/builds/triggering_builds.html#config-change-triggers
-  name: "{{ image_name(service_variant, edxapp_image_tag, edxapp_theme_tag) | replace(':', '-') }}"
+  name: "{{ internal_docker_registry }}/{{ project_name }}/{{ image_name(service_variant, edxapp_image_tag, edxapp_theme_tag) }}"
```
jmaupetit: This should not be changed.

apps/edxapp/templates/services/cms/job_01_internationalization.yml.j2
```diff
@@ -27,7 +27,7 @@ spec:
         - name: edxapp-cms-internationalization-{{ job_stamp }}
           # We point to a local registry image build for this "live" image (see
           # ImageStream and BuildConfig templates)
-          image: "{{ internal_docker_registry }}/{{ project_name }}/edxapp-cms:{{ edxapp_image_tag }}-live"
+          image: "{{ internal_docker_registry }}/{{ project_name }}/{{ image_name(service_variant, edxapp_image_tag, edxapp_theme_tag) }}"
```
jmaupetit: `service_variant` variable is not defined here, should be:

```suggestion
          image: "{{ internal_docker_registry }}/{{ project_name }}/{{ image_name('cms', edxapp_image_tag, edxapp_theme_tag) }}"
```

jmaupetit: The same applies for all other job templates modifications.

lunika (approved):

PR #391, opened 2019-11-05
🐛(marsha) test if sentry is defined before setting in secret

lunika:
## Purpose

Using sentry is not mandatory and using a default value in the secret is
wrong because the sdk will warn us that the dsn is incorrect. The good
way is to declare it in the secret only it was set in marsha's vault.

## Proposal

- [x] set sentry in marsha secret only if present in the vault

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Upgrade OpenShift to `0.10.0`
 - Upgrade `yq` ti `2.8.1`
 
+### Fixed
+
+- Set sentry in marsha secret only if present in the vault
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Upgrade OpenShift to `0.10.0`
 - Upgrade `yq` ti `2.8.1`
 
+### Fixed
+
+- set sentry in marsha secret only if present in the vault
```
sampaccoud: We usually start CHANGELOG lines by a capital letter.

sampaccoud (approved):

jmaupetit (approved):

PR #393, opened 2019-11-14
manage conditional volumes

lunika:
## Purpose

Creating volume can be conditioned by the env used when the application is
deployed. Templates can be empty and the creation will fail. We have to
filter volume templates to remove empty template and then create non
empty volumes.

## Proposal

- [x] filter empty volume templates
- [x] ignore volumes in marsha's nginx DC if not in trashable env

### DIFF
--- a/apps/marsha/templates/services/nginx/configs/marsha.conf.j2
+++ b/apps/marsha/templates/services/nginx/configs/marsha.conf.j2
@@ -29,8 +29,10 @@ server {
     try_files $uri @proxy_to_marsha_app;
   }
 
+  {% if env_type in trashable_env_types %}
   location ~ ^/static/(?P<file>.*) {
     root /data/static/marsha;
     try_files /$file =404;
   }
+  {% endif %}
 }
--- a/apps/marsha/templates/services/nginx/dc.yml.j2
+++ b/apps/marsha/templates/services/nginx/dc.yml.j2
@@ -42,16 +42,18 @@ spec:
             - mountPath: /etc/nginx/conf.d
               name: marsha-v-nginx
               readOnly: true
+{% if env_type in trashable_env_types %}
             - mountPath: /data/media/marsha
               name: marsha-v-media
               readOnly: true
             - mountPath: /data/static/marsha
               name: marsha-v-static
               readOnly: true
-            {% if activate_http_basic_auth or marsha_activate_http_basic_auth -%}
+{% endif %}
+{% if activate_http_basic_auth or marsha_activate_http_basic_auth %}
             - mountPath: "{{ http_basic_auth_user_file | dirname }}"
               name: marsha-htpasswd
-            {% endif %}
+{% endif %}
 
           livenessProbe:
             httpGet:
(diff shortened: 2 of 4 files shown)


tasks/create_app_volumes.yml
```diff
@@ -6,10 +6,24 @@
...
+  set_fact:
+    path: "{{ item }}"
+  register: filtered_volumes
+  with_items: "{{ app.volumes }}"
+  when: app.volumes is defined and lookup('template', item) | length > 1
+  tags: deploy
+
+- name: Update volume templates list for this app
+  set_fact:
+    volumes: "{{ filtered_volumes | json_query('results[*].ansible_facts.path') | list }}"
```
sampaccoud: I didn't understand why you can't use `filtered_volumes` directly?

apps/marsha/templates/services/nginx/dc.yml.j2
```diff
@@ -69,12 +71,14 @@ spec:
...
           configMap:
             name: marsha-nginx-{{ deployment_stamp }}
+{% if env_type in trashable_env_types %}
         - name: marsha-v-media
           persistentVolumeClaim:
             claimName: marsha-pvc-media
         - name: marsha-v-static
           persistentVolumeClaim:
             claimName: marsha-pvc-static
+{% if env_type in trashable_env_types %}
```
sampaccoud: maybe re-align below's "ifs"

apps/marsha/templates/services/nginx/dc.yml.j2
```diff
@@ -42,12 +42,14 @@ spec:
...
               name: marsha-v-nginx
               readOnly: true
+{% if env_type in trashable_env_types %}
             - mountPath: /data/media/marsha
               name: marsha-v-media
               readOnly: true
             - mountPath: /data/static/marsha
               name: marsha-v-static
               readOnly: true
+{% endif %}
```
sampaccoud, line 52: maybe re-align below's "ifs"

sampaccoud:
Actually you should also remove the nginx rule here: https://github.com/openfun/arnold/blob/master/apps/marsha/templates/services/nginx/configs/marsha.conf.j2#L32

apps/marsha/templates/services/nginx/dc.yml.j2
jmaupetit: Should be `endif` instead to close the block.

lunika:
> Actually you should also remove the nginx rule here: https://github.com/openfun/arnold/blob/master/apps/marsha/templates/services/nginx/configs/marsha.conf.j2#L32

Done

apps/marsha/templates/services/nginx/configs/marsha.conf.j2
```diff
@@ -29,8 +29,10 @@ server {
     try_files $uri @proxy_to_marsha_app;
   }
 
+  {% if env_type in trashable_env_types %}
   location ~ ^/static/(?P<file>.*) {
     root /data/static/marsha;
     try_files /$file =404;
   }
+  {% endif %}
```
jmaupetit, line 37: :1st_place_medal:

jmaupetit (approved):
Don't forget the CHANGELOG!

sampaccoud (approved):

PR #394, opened 2019-11-14
♻️(richie) allow deactivating static/media volumes

sampaccoud:
## Purpose

For most of our projects, we use AWS S3 & CloudFront to store and serve our static and media files. In this case, creating volumes in OpenShift is useless. 

## Proposal

I propose to allow setting volumes sizes to 0 to deactivate them.

For the moment, I propose to keep creating volumes by default because it is simpler and works out of the box. Also, relying on AWS requires a Terraform project that is not in richie's sandbox yet.

### DIFF
--- a/apps/richie/templates/services/app/job_01_collectstatic.yml.j2
+++ b/apps/richie/templates/services/app/job_01_collectstatic.yml.j2
@@ -26,6 +26,7 @@ spec:
       imagePullSecrets:
         - name: "{{ richie_image_pull_secret_name }}"
 {% endif %}
+      restartPolicy: Never
       containers:
         - name: richie-collectstatic
           image: "{{ richie_image_name }}:{{ richie_image_tag }}"
@@ -50,6 +51,7 @@ spec:
             - "bash"
             - "-c"
             - python manage.py collectstatic --noinput
+{% if richie_should_activate_static_volume %}
           volumeMounts:
             - mountPath: /data/static
               name: richie-v-static
@@ -57,4 +59,4 @@ spec:
         - name: richie-v-static
           persistentVolumeClaim:
             claimName: richie-pvc-static
-      restartPolicy: Never
+{% endif %}
--- a/apps/richie/templates/services/nginx/dc.yml.j2
+++ b/apps/richie/templates/services/nginx/dc.yml.j2
@@ -42,16 +42,20 @@ spec:
             - mountPath: /etc/nginx/conf.d
               name: richie-v-nginx
               readOnly: true
+{% if richie_should_activate_media_volume %}
             - mountPath: /data/media/richie
               name: richie-v-media
               readOnly: true
+{% endif %}
+{% if richie_should_activate_static_volume %}
             - mountPath: /data/static/richie
               name: richie-v-static
               readOnly: true
-            {% if activate_http_basic_auth or richie_activate_http_basic_auth -%}
+{% endif %}
+{% if activate_http_basic_auth or richie_activate_http_basic_auth -%}
             - mountPath: "{{ http_basic_auth_user_file | dirname }}"
               name: richie-htpasswd
-            {% endif %}
+{% endif %}
 
           livenessProbe:
             httpGet:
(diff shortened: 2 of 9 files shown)


apps/richie/vars/all/main.yml
```diff
@@ -58,5 +58,6 @@ richie_extra_dependencies:
 richie_activate_http_basic_auth: false
 
 # -- volumes
+# volume sizes with power of 2 suffices (Ki, Mi, Gi, Ti...) e.g 2Gi and 0 for no volume.
```
jmaupetit, line 61:
```suggestion
# volume sizes with power of 3 suffices (Ki, Mi, Gi, Ti...) e.g 2Gi and 0 for no volume.
```

apps/marsha/templates/services/nginx/dc.yml.j2
```diff
@@ -75,8 +75,8 @@ spec:
         - name: marsha-v-static
           persistentVolumeClaim:
             claimName: marsha-pvc-static
-        {% if activate_http_basic_auth or marsha_activate_http_basic_auth -%}
+{% if activate_http_basic_auth or marsha_activate_http_basic_auth -%}
```
lunika: I'm already doing it in my PR

lunika (commented):
For richie creating the volumes does not depend on the environment used ?

apps/richie/vars/all/main.yml
sampaccoud: Ki = 2^10 
Mi = 2^20
Gi = 2^30
Ti = 2^40

sampaccoud:
> For richie creating the volumes does not depend on the environment used ?

@lunika in trashable envs we probably won't use volumes but it is weird to force it in Arnold when in fact not requiring the volumes is something that is first set in Django settings. So it makes more sense to configure both Django and OpenShift templates in Matsuo to use volumes or not.

lunika:
> > For richie creating the volumes does not depend on the environment used ?
> 
> @lunika in trashable envs we probably won't use volumes but it is weird to force it in Arnold when in fact not requiring the volumes is something that is first set in Django settings. So it makes more sense to configure both Django and OpenShift templates in Matsuo to use volumes or not.

In that case the default value should be 0 and you explicitly activate them when you need them. WDYT ?

sampaccoud:
@lunika The problem with making this the default behavior is that we should also provide the Terraform in richie... and it adds a dependance to AWS when some people may want to use something else...

apps/marsha/templates/services/nginx/dc.yml.j2
sampaccoud: yes you're right it's a mistake. I will anyway need to rebase on yours.

apps/richie/vars/all/main.yml
jmaupetit: TIL. My mistake. I've double-misinterpreted your comment :sweat_smile:

jmaupetit (commented):
So you prefer not indenting tags?

sampaccoud:
> So you prefer not indenting tags?

I have no preference but I had the feeling that it was the most current/recent practice in Arnold?

jmaupetit (approved):
You'll have to wait for #393 to be merged!

And also: **don't forget the CHANGELOG!**

apps/richie/vars/all/main.yml
sampaccoud, line 61: @jmaupetit I'm having second thoughts, isn't it better to have separate variables?
```yml
- richie_should_activate_media_volume: true
- richie_should_activate_static_volume: true
```

jmaupetit: Indeed. It would be more explicit.

sampaccoud: @jmaupetit Done and ready for further review.

apps/richie/templates/services/app/job_01_collectstatic.yml.j2
```diff
@@ -50,11 +51,12 @@ spec:
             - "bash"
             - "-c"
             - python manage.py collectstatic --noinput
+{% if richie_should_activate_media_volume %}
```
sampaccoud: oh shit this is wrong

apps/richie/templates/services/nginx/dc.yml.j2
```diff
@@ -42,16 +42,20 @@ spec:
             - mountPath: /etc/nginx/conf.d
               name: richie-v-nginx
               readOnly: true
+{% if richie_should_activate_media_volume %}
             - mountPath: /data/media/richie
               name: richie-v-media
               readOnly: true
+{% endif %}
+{% if richie_should_activate_media_volume %}
```
sampaccoud: wrong also

lunika (approved):

PR #396, opened 2019-11-25
Configure MongoDB to use a list IPs, a replicaSet and read_preference settings

lunika:
## Purpose

We want to configure all the applications to use a MongoDB replicaset and also configure the read_preference setting. To do that all the applications should be configured with a list of MongoDB IPs and also use a setting to configure the replicaSet and the read_preference.

## Proposal

- [x] transform MongoDB ip in a list of IPs
- [x] configure learninglocker to use MongoDB replicaset
- [x] configure forum to use MongoDB replicaset
- [x] configure edxapp to use MongoDB replicaset

/!\ This PR is a BC break

### DIFF
--- a/apps/edxapp/templates/services/cms/macros/mongodb_host.yml.j2
+++ b/apps/edxapp/templates/services/cms/macros/mongodb_host.yml.j2
@@ -1,0 +1,16 @@
+{# compute mongodb host #}
+{%- macro mongodb_host(env_type, trashable_env_types, deployment_stamp, endpoint_mongodb_ips, edxapp_mongodb_port, edxapp_mongodb_host) %}
+    {%- if env_type in trashable_env_types -%}
+        {{ edxapp_mongodb_host }}-{{ deployment_stamp }}:{{ edxapp_mongodb_port }}
+    {%- else -%}
+        {%- set mongodb_service_name = "%s-%s" | format(edxapp_mongodb_host, deployment_stamp) -%}
+        {%- set ns = namespace(hosts="") -%}
+        {%- for ip in endpoint_mongodb_ips -%}
+            {%- set ns.hosts = "%s%s:%s" | format(ns.hosts, mongodb_service_name, edxapp_mongodb_port | int + loop.index0) -%}
+            {%- if loop.index < endpoint_mongodb_ips | length -%}
+                {%- set ns.hosts = "%s," | format(ns.hosts) -%}
+            {%- endif -%}
+        {%- endfor -%}
+        {{ ns.hosts }}        
+    {%- endif -%}
+{%- endmacro %}
--- a/apps/forum/vars/all/main.yml
+++ b/apps/forum/vars/all/main.yml
@@ -21,4 +21,7 @@ forum_mongodb_version: "3.2"
 forum_mongodb_image_name: "centos/mongodb-32-centos7"
 forum_mongodb_image_tag: "3.2"
 forum_mongodb_port: 27017
+forum_mongodb_host: "forum-mongodb"
+forum_mongodb_replicaset: ""
+forum_mongodb_read_preference: ""
 forum_mongodb_secret_name: "forum-mongo-{{ forum_vault_checksum | default('undefined_forum_vault_checksum') }}"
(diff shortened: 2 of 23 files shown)


apps/edxapp/templates/services/cms/_dc_base.yml.j2
```diff
@@ -83,6 +84,16 @@ spec:
           value: {{ service_variant }}
         - name: DJANGO_SETTINGS_MODULE
           value: {{ service_variant }}.envs.fun.docker_run
+        - name: MONGODB_HOST
+          value:  "{{ mongodb_host(env_type, trashable_env_types, deployment_stamp, endpoint_mongodb_ips, edxapp_mongodb_port, edxapp_mongodb_host) }}"
+{% if edxapp_mongodb_replicaset is defined and edxapp_mongodb_replicaset | length > 1 %}
+        - name: MONGODB_REPLICASET
+          value: {{ edxapp_mongodb_replicaset }}
+{% endif %}
+{% if edxapp_mongodb_read_preference is defined and edxapp_mongodb_read_preference | length > 1 %}
```
sampaccoud: put this inside the above if as we don't want it when there is no replicaset?


Actually I don't understand why you need to set these variables here since you already set them in: apps/edxapp/templates/services/cms/configs/settings.yml.j2

apps/forum/vars/all/main.yml
```diff
@@ -21,4 +21,7 @@ forum_mongodb_version: "3.2"
 forum_mongodb_image_name: "centos/mongodb-32-centos7"
 forum_mongodb_image_tag: "3.2"
 forum_mongodb_port: 27017
+forum_mongodb_host: "forum-mongo"
+forum_mongodb_replicaset: ""
```
sampaccoud, line 25: is none better than ""?

apps/edxapp/templates/services/cms/macros/mongodb_host.yml.j2
```diff
@@ -0,0 +1,16 @@
+{# compute mongodb host #}
+{%- macro mongodb_host(env_type, trashable_env_types, deployment_stamp, endpoint_mongodb_ips, edxapp_mongodb_port, edxapp_mongodb_host) %}
+    {%- if env_type in trashable_env_types -%}
+        {{ edxapp_mongodb_host }}-{{ deployment_stamp }}:{{ edxapp_mongodb_port }}
+    {%- else -%}
+        {%- set mondogdb_service_name = "%s-%s" | format(edxapp_mongodb_host, deployment_stamp) -%}
```
jmaupetit:
```suggestion
        {%- set mongodb_service_name = "%s-%s" | format(edxapp_mongodb_host, deployment_stamp) -%}
```

apps/edxapp/templates/services/cms/macros/mongodb_host.yml.j2
```diff
@@ -0,0 +1,16 @@
...
+        {%- set ns = namespace(hosts="") -%}
+        {%- for ip in endpoint_mongodb_ips -%}
+            {%- set ns.hosts = "%s%s:%s" | format(ns.hosts, mondogdb_service_name, edxapp_mongodb_port | int + loop.index0) -%}
```
jmaupetit: That double typo saved you :wink: 

```suggestion
            {%- set ns.hosts = "%s%s:%s" | format(ns.hosts, mongodb_service_name, edxapp_mongodb_port | int + loop.index0) -%}
```

jmaupetit (commented):
The `mongodb_host` macros appear too convoluted and less maintainable than a custom filter. I would prefer testing and maintaining python code for such complicated configuration.

apps/edxapp/templates/services/cms/_dc_base.yml.j2
lunika: oh yes I have to remove them fron the settings

sampaccoud: but they are better in the settings.yml file since all the other settings are there... ?

lunika: ok

lunika: done

sampaccoud (approved):

PR #400, opened 2019-11-26
♻️(registry) use local lookupPolicy

lunika:
## Purpose

ImageStream can use first the local registry and then fetch on
docker hub if the image is not present. To do that the local
lookupPolicy must be set to true and there is no need anymore to refer
to the internal docker registry in the image name used in every
DeploymentConfig or jobs.

## Proposal

- [x] configure ImageStream with local lookupPolicy set to true


sampaccoud (approved):

lunika:
Using local lookupPolicy doesn't work on the CI. I suggest to not merge it, we don't need it on our OC cluster for now.

PR #402, opened 2019-11-28
✨(app) create a redis-sentinel application

lunika:
## Purpose

we choose to use the redis-operator
(https://github.com/spotahome/redis-operator) to deploy a redis
configured with sentinel. Only one OpenShift object is used and deployed as a
Service.

## Proposal

- [x] implement the redis-sentinel app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Create a new app redis-sentinel. This app uses the redis-operator 
+  (https://github.com/spotahome/redis-operator) and it must be installed
+  on your OpenShift cluster before deploying.
+
 ## [4.0.0] - 2019-11-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   (https://github.com/spotahome/redis-operator) and it must be installed
   on your OpenShift cluster before deploying.
 
+### Removed
+
+- Probes on edxapp workers. Existing probes are not compatible with redis sentinel.
+  To use them with redis sentinel, we must use a Django management command or Celery
+  using Django settings, but doing so is too slow to be considered as a relevant alternative.
+  This removal is temporary, a new solution will be developped soon.
+
 ## [4.0.0] - 2019-11-26
 
 ### Added
--- a/apps/redis-sentinel/templates/services/app/svc.yml.j2
+++ b/apps/redis-sentinel/templates/services/app/svc.yml.j2
@@ -1,0 +1,56 @@
+{# 
+  Disclaimer: This template is not relative to an OpenShift Service object but to a RedisFailover object.
+  RedisFailover is part of the redis-operator (https://github.com/spotahome/redis-operator) and must be
+  deploy at the same time of other OpenShift Service, this is why we decided to created in this file.
+#}
+apiVersion: databases.spotahome.com/v1
+kind: RedisFailover
+metadata:
+  labels:
+    app: redis-sentinel
+    service: app
+    version: RedisFailover
+  # Nota bene: the name of the service should be the host name in edxapp configuration,
+  # prefixed by "rfs-", e.g. "rfs-redis-sentinel"
+  name: redis-sentinel
+  namespace: "{{ project_name }}"
+spec:
+  sentinel:
+    replicas: {{ redis_sentinel_sentinel_replicas }}
+{% if redis_sentinel_sentinel_custom_config | length > 0 %}
+    customConfig:
+{% for config in redis_sentinel_sentinel_custom_config %}
+      - "{{ config }}"
+{% endfor %}
+{% endif %}
+    securityContext:
+      runAsUser: {{ redis_sentinel_user_id }}
+      runAsGroup: {{ redis_sentinel_group_id }}
+      fsGroup: {{ redis_sentinel_fsgroup_id }}
+  redis:
+    replicas: {{ redis_sentinel_redis_replicas }}
+{% if redis_sentinel_redis_custom_config | length > 0 %}
+    customConfig:
+{% for config in redis_sentinel_redis_custom_config %}
+      - "{{ config }}"
+{% endfor %}
+{% endif %}
+    storage:
+      keepAfterDeletion: true
+      persistentVolumeClaim:
+        metadata:
+          name: redis-sentinel-pvc-data
+          namespace: "{{ project_name }}"
+          labels:
+            app: redis-sentinel
+            version: RedisFailover
+        spec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: {{ redis_sentinel_redis_data_volume_size }}
+    securityContext:
+      runAsUser: {{ redis_sentinel_user_id }}
+      runAsGroup: {{ redis_sentinel_group_id }}
+      fsGroup: {{ redis_sentinel_fsgroup_id }}
--- a/apps/redis-sentinel/vars/all/main.yml
+++ b/apps/redis-sentinel/vars/all/main.yml
@@ -1,0 +1,26 @@
+# The following settings can be found in the project's annotation.
+# To retrieve them run the following command:
+#   $ oc project {{ project_name }} -o yaml
+# then copy values of the `openshift.io/sa.scc.supplemental-groups` annotation
+redis_sentinel_user_id: null
+redis_sentinel_group_id: null
+redis_sentinel_fsgroup_id: null
+
+## Sentinel configuration
+# Number of sentinel replicas 
+redis_sentinel_sentinel_replicas: 3
+# If you need custom sentinel configuration you should add
+# them in this list with one configuration per item, e.g.:
+#   - "sentinel down-after-milliseconds mymaster 2000"
+#   - "down-after-milliseconds 2000"
+redis_sentinel_sentinel_custom_config: []
+
+## Redis configuration
+# Number of redis replicas
+redis_sentinel_redis_replicas: 3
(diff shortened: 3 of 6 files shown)


apps/redis-sentinel/vars/settings.yml
```diff
@@ -0,0 +1 @@
+is_blue_green_compatible: False
```
jmaupetit:
```suggestion
is_blue_green_compatible: false
```

apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -0,0 +1,55 @@
+{# 
+This template is not relative to an OpenShift Service object but to a RedisFailover one.
+RedisFailover is part of redis-operator (https://github.com/spotahome/redis-operator) ans must be
+deploy at the same time of other OpenShift Service, this is why we decided to created in this file.
```
jmaupetit: Can you indent this comment in the block?

apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -0,0 +1,55 @@
+{# 
+This template is not relative to an OpenShift Service object but to a RedisFailover one.
```
jmaupetit:
```suggestion
Disclaimer: this template is not relative to an OpenShift Service object but to a RedisFailover object.
```

apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -0,0 +1,55 @@
...
+RedisFailover is part of redis-operator (https://github.com/spotahome/redis-operator) ans must be
```
jmaupetit:
```suggestion
RedisFailover is part of the redis-operator (https://github.com/spotahome/redis-operator) and must be
```

apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -0,0 +1,55 @@
...
+deploy at the same time of other OpenShift Service, this is why we decided to created in this file.
+#}
+apiVersion: databases.spotahome.com/v1
+kind: RedisFailover
+metadata:
+  labels:
+    app: redis-sentinel
+    service: app
+    version: RedisFailover
+  # name of the service should be host name in edxapp configuration prefixed by rfs- : rfs-redis-sentinel
```
jmaupetit:
```suggestion
  # Nota bene: the name of the service should be the host name in edxapp configuration,
  # prefixed by "rfs-", e.g. "rfs-redis-sentinel"
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -0,0 +1,26 @@
+# This setting can be found in the project annotation
```
jmaupetit:
```suggestion
# The following settings can be found in the project's annotation.
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -0,0 +1,26 @@
...
+# to retrieve them run oc project {{ project_name }} -o yaml
```
jmaupetit:
```suggestion
# To retrieve them run the following command:
#   $ oc project {{ project_name }} -o yaml
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -0,0 +1,26 @@
...
+# then copy the value of the annotation openshift.io/sa.scc.supplemental-groups
```
jmaupetit:
```suggestion
# then copy values of the `openshift.io/sa.scc.supplemental-groups` annotation
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -0,0 +1,26 @@
...
+#   - "sentinel down-after-milliseconds mymaster 2000"
+#   - "down-after-milliseconds 2000"
+redis_sentinel_sentinel_custom_config: []
+
+## Redis configuration
+# Number of redis replicas
+redis_sentinel_redis_replicas: 3
+redis_sentinel_redis_data_volume_size: 2Gi
+# If you need custom redis configuration you should add
+# them in this list. One configuration is a new entry
```
jmaupetit, line 23:
```suggestion
# them in this list with one configuration per item, e.g.:
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -0,0 +1,26 @@
...
+# then copy the value of the annotation openshift.io/sa.scc.supplemental-groups
+redis_sentinel_user_id: null
+redis_sentinel_group_id: null
+redis_sentinel_fsgroup_id: null
+
+## Sentinel configuration
+# Number of sentinel replicas 
+redis_sentinel_sentinel_replicas: 3
+# If you need custom sentinel configuration you should add
+# them in this list. One configuration is a new entry
```
jmaupetit:
```suggestion
# them in this list with one configuration per item, e.g.:
```

jmaupetit (approved):
Great work!

jmaupetit:
Even tho, we will need a new PR allowing to use Redis sentinel in `edxapp` (see worker probes).

jmaupetit (changes requested):
In fact we need to fix `edxapp` workers probes that would be broken if we switch to redis sentinel. See https://github.com/openfun/arnold/blob/master/apps/edxapp/templates/services/cms/_dc_base.yml.j2#L96

lunika:
> In fact we need to fix `edxapp` workers probes that would be broken if we switch to redis sentinel. See https://github.com/openfun/arnold/blob/master/apps/edxapp/templates/services/cms/_dc_base.yml.j2#L96

This should be done in an other PR ? This PR just target the creation of a new app

jmaupetit:
> This PR just target the creation of a new app

Indeed, but it's unusable as is.

lunika:
Ok so the only solution I found for now is to transform the command with this:

```
celery -A lms.celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)
```

WDYT @sampaccoud @jmaupetit

sampaccoud:
@lunika I would say that if we want the edxapp healthcheck to validate that redis is working it should be done via the edxapp healthcheck endpoint? But I'm not really aware of existing probes :man_shrugging:

lunika:
> @lunika I would say that if we want the edxapp healthcheck to validate that redis is working it should be done via the edxapp healthcheck endpoint? But I'm not really aware of existing probes man_shrugging

This probe is not used by the wsgi application but by every worker deployed so the healthcheck endpoint is not usable.

sampaccoud:
Then we should add a small management command in edxapp that checks the redis connection (with or without sentinel depending on the settings the same way as Celery or django-redis-session are connecting to redis...). I would make a simple command that writes a value to redis and reads it again... 
Actually it could also be the same endpoint for the wsgi application and the worker pods... WDYT?

jmaupetit:
Loading a management command with the `edx`stack is too heavy for a health check. That was our first attempt and it proved to be unusable.

jmaupetit:
> Ok so the only solution I found for now is to transform the command with this:
> 
> ```
> celery -A lms.celery inspect ping -d celery@{{ celery_worker.name }}.%$(hostname)
> ```
> 
> WDYT @sampaccoud @jmaupetit

Have you tried it with `time` in a worker pod to evaluate its responsiveness.

lunika:
``` bash
$ time celery -A lms.celery inspect ping -d celery@lms-default.%$(hostname)
-> celery@lms-default.%edxapp-lms-default-worker-d-190717-13h37m29s-2-sf9qm: OK
        pong

real	0m1.211s
user	0m0.992s
sys	0m0.620s
```

jmaupetit:
Did you run it multiple times (at least 3) to check execution time "standard deviation"? IMO 1.2s is not acceptable for such probe.

lunika:
> Did you run it multiple times (at least 3) to check execution time "standard deviation"? IMO 1.2s is not acceptable for such probe.

Yes I run it multiple time and this is one the best time I have.

jmaupetit:
So it's a dead end.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   (https://github.com/spotahome/redis-operator) and it must be installed
   on your OpenShift cluster before deploying.
 
+### Removed
+
+- probes on edxapp workers. They are too slow to be run
```
jmaupetit: This is not true: the real problem is that existing probes are not compatible with redis sentinel and alternative probes based on Django management commands or Celery using Django settings are too slow to be considered as a relevant alternative. You must add that this is a temporary removal until we implement a new efficient probe.

CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
 
+### Added
+
+- Create a new app redis-sentinel. This app uses the redis-operator 
+  (https://github.com/spotahome/redis-operator) and it must be installed
+  on your OpenShift cluster before deploying.
+
+### Removed
+
+- probes on edxapp workers. Existing probes are not compatible with redis sentinel.
```
```suggestion
- Probes on edxapp workers. Existing probes are not compatible with redis sentinel.
```

CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
+  To use them with redis sentinel, we must use a Django management command or Celery
+  using Django settings, doing this is too slow to be considered as a relevant alternative.
```
```suggestion
  using Django settings, but doing so is too slow to be considered as a relevant alternative.
```

CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
+  This removal is temporary, a new solution must be developed soon.
```
```suggestion
  This removal is temporary, a new solution will be developed soon.
```

jmaupetit (approved):

PR #405, opened 2019-12-09
🐛(learninglocker) Inject XAPI secret as environment variable in XAPI DC

lunika:
## Purpose

A secret is created for the XAPI service but unused by the XAPI DC.
Without it, it is not possible to connect to MongoDB because the
credentials are in this secret and must be injected as environment
variable.

## Proposal

- [x] inject XAPI secret as environment variable in XAPI DC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
```
sampaccoud, line 8: in your commit message:
- s/A secret in created/A secret is created

sampaccoud (approved):

PR #407, opened 2019-12-09
🐛(tasks) fix redeploying apps that have no deployment config and release 4.1.2

sampaccoud:
## Purpose

The redis-sentinel app has no dc and works only by declaring an svc. The first deployment works perfectly, but redeployments fail because we endup expecting 0 pods running when there are in fact pods running (they have been started by the k8s operator and not by a dc created
by Arnold).

## Proposal

Instead of checking if the number of running pods is equal to the number of expected pods, we can check if the number of running pods is greater than the number of expected pods.


tasks/manage_app.yml
```diff
@@ -84,7 +84,7 @@
     field_selectors:
       - "status.phase=Running"
   register: pods
-  until: pods.resources is defined and ( pods.resources | length ) == ( expected_running_pods | int )
+  until: pods.resources is defined and ( pods.resources | length ) >= ( expected_running_pods | int )
```
lunika: we could also add a `when` instruction testing if `deployments | length > 0`

```
when: deployments |  length > 0
```

jmaupetit: Indeed, I think your solution is more reliable @lunika, but why not using:

```yaml
  when: expected_running_pods | int > 0
```

jmaupetit (commented):
Don't forget the CHANGELOG!

tasks/manage_app.yml
sampaccoud: Ok I did this. Checking on `expected_running_pods` is easier to understand as we have this variable at hand here.

sampaccoud:
Added the 4.1.2 fix release to this PR.

jmaupetit (approved):

sampaccoud:
@jmaupetit worse GTM gif I got from you!

tasks/manage_app.yml
lunika: > Indeed, I think your solution is more reliable @lunika, but why not using:
> 
> ```yaml
>   when: expected_running_pods | int > 0
> ```

Yes, you are a very clever man

PR #408, opened 2019-12-10
✨(redis-sentinel) apply podAntiAffinity rules

lunika:
## Purpose

We don't want have two instances of the same component on the same
node. For this we can use a podAntiAffinity rule to exclude pod having
the same component value on the same node.

## Proposal

- [x] Apply PodAntiAffinity rules to redis-sentinel app

### DIFF
--- a/apps/redis-sentinel/templates/services/app/svc.yml.j2
+++ b/apps/redis-sentinel/templates/services/app/svc.yml.j2
@@ -17,6 +17,18 @@ metadata:
 spec:
   sentinel:
     replicas: {{ redis_sentinel_sentinel_replicas }}
+    affinity:
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                  - key: app.kubernetes.io/component
+                    operator: In
+                    values:
+                      - sentinel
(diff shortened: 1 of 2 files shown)


apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -29,6 +40,17 @@ spec:
...
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                  - key: app.kubernetes.io/component
+                    operator: In
+                    values:
+                      - redis
```
sampaccoud, line 54: if I look at what we did on edxapp, you are missing the most important line: "topologyKey: kubernetes.io/hostname"?

lunika: I don't think we have to use it, this topologykey is specific to the selector we use on the other DC (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#built-in-node-labels) here it's a custom label added by the operator. I don't know if there is a topologykey for this case.

lunika: You are right and I am wrong. I added the topologykey

sampaccoud (approved):

PR #411, opened 2019-12-17
🚸(apps) run the db_migrate job before starting application pods

sampaccoud:
## Purpose

If we run db_migrate after starting the new pod:

- the current pod has to be compatible with the state of the db before and after the migration.
- the new pod has to be compatible with the state of the database before and after a migration. This is not possible for a new database, and not easy for schema updates.

If we run db_migrate before starting a new pod, the new pod only has to be
compatible with the state of the database after the migration.

## Proposal

Tag all `db_migrate` jobs as `pre` jobs.

### DIFF
--- a/apps/edxapp/templates/services/cms/job_05_db_migrate.yml.j2
+++ b/apps/edxapp/templates/services/cms/job_05_db_migrate.yml.j2
@@ -6,10 +6,12 @@ metadata:
   namespace: "{{ project_name }}"
   labels:
     app: edxapp
+    deployment_stamp: "{{ deployment_stamp }}"
+    job_stamp: "{{ job_stamp }}"
+    # Jobs with the "pre" job type will be executed prior to deployments
+    job_type: "pre"
     service: cms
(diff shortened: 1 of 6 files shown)


lunika (approved):

jmaupetit (approved):

apps/edxapp/templates/services/cms/job_05_db_migrate.yml.j2
```diff
@@ -6,10 +6,12 @@ metadata:
   namespace: "{{ project_name }}"
   labels:
     app: edxapp
+    deployment_stamp: "{{ deployment_stamp }}"
+    job_stamp: "{{ job_stamp }}"
+    # Jobs with the "pre" job type will be executed prior to deployments
+    job_type: "pre"
```
jmaupetit, line 12: Should be conditioned to non-trashable environments.

jmaupetit (changes requested):

PR #424, opened 2020-01-07
🔧(edxapp) configure routing timeout

lunika:
## Purpose

some pages need more time to be computed and we have a timeout error. To
resolve this issue, the LMS and CMS nginx configuration must be adapted
by setting proxy_read_timeout value to a higher one. Also OpenShift
routes, using haproxy, also have a timeout we need to increase. We want
the same timeout everywhere so only one new setting is added at the
edxapp application level: edxapp_routing_timeout

## Proposal

- [x] add a new setting `edxapp_routing_timeout`
- [x] use the setting `edxapp_routing_timeout` in nginx and routes configurations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [4.3.1] - 2019-12-23
 
+### Added
+
+- Configure routing timeout in edxapp application
+
 ### Changed
 
 - Upgrade the learninglocker image to v5.2.2 and the xapi-service image to v2.9.10
--- a/apps/edxapp/templates/services/nginx/configs/cms.conf.j2
+++ b/apps/edxapp/templates/services/nginx/configs/cms.conf.j2
@@ -24,6 +24,8 @@ server {
   location @proxy_to_cms_app {
     proxy_set_header Host $http_host;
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [4.3.1] - 2019-12-23
 
+### Added
+
+- Configure routing timeout in edxapp application.
+  routes and all nginx configurations share the same setting: edxapp_routing_timeout
```
jmaupetit: This is too verbose for a CHANGELOG IMO.
```suggestion

```

jmaupetit (approved):

CHANGELOG.md
lunika: I decided to be in verbose mode this year.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [4.3.1] - 2019-12-23
 
+### Added
+
+- Configure routing timeout in edxapp application
```
jmaupetit, line 25: @lunika I didn't noticed by it seems that you should be in the unreleased section instead of `4.3.1`... :sweat_smile:

Issue #417, opened 2019-12-31
Remove services from the blue/green dance


sampaccoud:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Arnold supports blue/green deployments by deploying complete stacks of **services**, pods, jobs, endpoints, etc. and switching the current route to the service of the new stack.

To do this, services include a deployment stamp in their name. This has raised several issues because it is not possible to use service names in the settings of another app because the name will change as soon as the service is redeployed.

Concrete cases when this is a problem:
- to run FUN's `dogwood.3-fun` image, we use the `elasticsearch` cluster from the `forum` app. The problem was solved temporarily by retrieving dynamically the host name of `elasticsearch`:
    ```
    {% set
      forum_current_service =
        lookup(
          'k8s',
          namespace=project_name,
          kind='Route',
          label_selector='app=forum,route_prefix=current'
        ) | json_query('spec.to.name')
    %}

    {% set
      forum_current_deployment_stamp =
        lookup(
          'k8s',
          namespace=project_name,
          kind='Service',
          resource_name=forum_current_service
        ) | json_query('metadata.labels.deployment_stamp')
    %}
    ```
    This is very brittle because the connection will break if we redeploy the `forum`, and it can only be repaired by subsequently redeploying `edxapp` with a downtime >30 minutes for the `forum`.
- the `edxapp` application needs to connect to the `forum` application which is in a separate app, both being deployed with blue/green. In order to have a stable address for the `forum`, we had to expose the forum on a route. This would not be necessary if we could connect to the `forum` service directly. The above trick used to connect to `elasticsearch` could also be used here, but it would be brittle and could only be temporary.

**Describe the solution you'd like**
We should take services out of the blue/green dance.

Services would be created at project initialization, with a fixed name (e.g. `edxapp-lms`). Routes would be connected to their services once for all.

The switch playbook would patch each service's selector instead of patching the target service in the route.

Thanks to the selector, it should be possible to make the switch in 2 steps:

1. add the next (green) stack to the selector. After this step, the traffic would go both to the current and to the next stacks. 
2. remove the current (blue) stack from the selector. After this step, all the traffic would go to the new stack.

This two-steps switch presents to advantages:

- zero downtime (note that switching the target service on the route was creating a small interruption)
- possibility to do canary deployments

jmaupetit:
Is the two-steps switch required? Why not patching the service once for all to target the new stack?

sampaccoud:
[...]

Issue #253, opened 2019-02-18
Remove black magic from ConfigMap generation


jmaupetit:
## Purpose

ConfigMap objects generation is too convoluted: by default, we inject files configuration to a ConfigMap for every file in a `_configs` service template path, and, if there is an `_env.yml.j2` file in a service templates path, we now create a config map that will be considered as environment variables for a service. These conventions add too much complexity/fragility to objects definitions.

## Proposal

As we did for routes, I think we must explicitly switch to `cm_*.yml.j2` files definitions for every applications/services. A service should be considered as self-consistent: one should NOT read our playbooks to understand which (and how) objects will be created for an application.

Issue #10, opened 2018-05-23
Remove unused '.vault_pass.sh'


jmaupetit:
## Purpose

`.vault_pass.sh` was used to ease ansible vault decryption from OpenShift. It is no longer required.

## Proposal

Remove this file from the repository :smile:

PR #427, opened 2020-01-08
✨(playbook) introduce static services

madmatah:
## Purpose

As explained in #417 , the current blue/green deployment mechanism adds a suffix (deployment stamp) to the name of the services exposed by applications.

This has raised several issues because it is not possible to use service names in the settings of another app because the name will change as soon as the service is redeployed.

## Proposal

Services with suffixes are helpful because they allow components of a deployed application to communicate with each other while also ensuring that the application stack is consistent within the same version. 

I propose to keep them, otherwise each component of an application must be aware of their current state (green or blue) to communicate with other components. And it's hard to do that without any reconfiguration during a switch between blue/green version. We don't want that.

In addition to the current mechanism, I propose to introduce "static services".
Each static service is declined in 3 versions (service-previous, service-current and service-next) that points to the right version of a deployment.

It can be used by another application because its name doesn't change.

The switch (or any deployment related task) is now patching static services instead of routes.
Routes are now pointing to the corresponding service (e.g., route _A-previous_ points to _service A-previous_).

A static service can of course exist without any route.

There is a new playbook `init_services.yml` that initialize static services.
Static services are defined in files matching `static-svc*.yml`

## TODO

- [x] POC with richie
- [x] Improve switch and rollback playbooks to avoid downtime (addressed in #434)
- [x] Inventory of the nice-to-have static services per application

Per application :
- [x] Update route templates
- [x] Add static service for each route
- [x] Add other static service if needed and patch applications using them.


create_services.yml
```diff
@@ -0,0 +1,20 @@
...
+      debug: msg="==== Starting create_services playbook ===="
+      tags: deploy
+
+    - import_tasks: tasks/set_vars.yml
+
+    - include_tasks: tasks/run_tasks_for_apps.yml
+      vars:
+        tasks:
+          - get_objects_for_app
+          - create_static_services
```
lunika, line 18: you create only static services, you should maybe rename this file in `create_static_services` ?

tasks/create_static_services.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Create static routes for services of an app
+
+- name: "Make sure static service exists for the '{{ app.name }}' app with all prefixes"
+  include_tasks: tasks/deploy_patch_static_service.yml
+  loop: "{{ blue_green_route_prefixes }}"
```
lunika: we should maybe rename this variable in `blue_green_prefixes` ?

madmatah: :+1:

create_services.yml
madmatah: Thanks for your feedbacks !

I tried to do that in a first place, but it triggered an error.
The only reason I see is because there is already a file named `create_static_services.yml` in the tasks folder.

tasks/create_static_services.yml
jmaupetit: :thinking: not sure as the prefixes are targeting routes?

madmatah: Now these prefixes are used for routes _and_ static services (e.g. route `app-next` points to service `app-next`)

madmatah:
FYI I've not removed the forum route.

I propose to do that later to ease the migration process.

jmaupetit (commented):
I would have preferred substituting existing SVCs by static ones without introducing a new semantic / object. Do you think we must keep B/G SVCs?

madmatah:
> I would have preferred substituting existing SVCs by static ones without introducing a new semantic / object. Do you think we must keep B/G SVCs?

B/G SVCs (with stamp) are essential because they allow components inside a stack to communicate with each other.

Our requirements are : 

1. We need to expose public HTTP(s) endpoints with consistent URLs ( -> **routes**)
2. We need to expose tcp or http services (internally) with consistent fqdn (-> **static-svc**)
3. We need to switch green/blue stack without downtime ( -> **static-svc** )
4. We need to provide a way for components inside a same stack to communicate with each other without any reconfiguration/restart. ( -> **bg svc**)

I didn't find a way to address this with less than 3 object types.

jmaupetit:
From my experience (I might be wrong), SVCs (B/G) almost never change. Can you provide a concrete example for the 4. point of your list?

madmatah:
> From my experience (I might be wrong), SVCs (B/G) almost never change. Can you provide a concrete example for the 4. point of your list?

For example, in **marsha** :

A reverse proxy (_marsha-nginx_) needs to communicate with its gunicorn backends (_marsha-app_) via a service (_marsha-app_).

In the nginx configuration, we have : 

```
upstream marsha-backend {
  server marsha-app-{{ deployment_stamp }}:{{ marsha_django_port }} fail_timeout=0;
}
```

The deployment stamp is useful here : it guarantees that we communicate with the gunicorn backends in the same stack, without being conscious that we are the current, next or previous version of the app.

This parameter cannot be dynamic in nginx and it would require a reconfiguration/restart of the nginx server if we want to change it after a switch.

tasks/delete_app_objects_kind.yml
```diff
@@ -7,6 +7,7 @@
     label_selectors:
       - app={{ app.name }}
       - deployment_stamp={{ targeted_deployment_stamp }}
+      - removable!=no
```
lunika, line 10: does this match object not having `removable` defined ?

madmatah: Yes, I tested it.

You can try it on a running openshift cluster with :
`oc get po -l 'test!=yolo'`

I did not find any reference of this specific behavior in the official documentation about label inequality, but you can get more information [here (Equality-based selection)](http://docs.containership.io/en/articles/1684800-labels-selectors).

apps/edxapp/templates/services/nginx/route_cms.yml.j2
```diff
@@ -25,4 +24,4 @@ spec:
     targetPort: "{{ edxapp_nginx_cms_port }}-tcp"
   to:
     kind: Service
-    name: "edxapp-nginx-{{ deployment_stamp }}"
+    name: "edxapp-nginx-{{ prefix }}"
```
sampaccoud, line 30: more like a suffix?

[edit] after reading the rest of the code I think it should actually be called `stack`.

switch.yml
```diff
@@ -1,9 +1,9 @@
 ---
-# This playbook switch all routes:
-# 1. point the "previous" route to the stack that was pointed by the "current" route,
-# 2. point the "current" route to the stack that was pointed by the "next" route,
-# 3. delete the stack that was pointed by the "previous" route,
-# 4. patch the next route to the init stamp
+# This playbook switch all static services:
```
sampaccoud:
```suggestion
# This playbook switches all static services:
```

tasks/switch_static_services.yml
```diff
@@ -1,12 +1,12 @@
 ---
-# Patch app dest routes with src stack
+# Patch app dest static services with src stack
 
-- include_tasks: deploy_get_stamp_from_route.yml
+- include_tasks: deploy_get_stamp_from_static_service.yml
```
sampaccoud, line 4: actually, I don't understand why the name of this task starts by `deploy`. Why not call it `get_stamp_from_static_service`?

apps/edxapp/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
sampaccoud: shouldn't you remove the existing svc when creating these static services?

tasks/get_objects_for_app.yml
```diff
@@ -18,6 +18,7 @@
     deployments: "{{ templates | map('regex_search', '.*/dc.*\\.yml\\.j2$') | select('string') | list }}"
     endpoints: "{{ templates | map('regex_search', '.*/ep.*\\.yml\\.j2$') | select('string') | list }}"
     services: "{{ templates | map('regex_search', '.*/svc.*\\.yml\\.j2$') | select('string') | list }}"
+    static_services: "{{ templates | map('regex_search', '.*/static-svc.*\\.yml\\.j2$') | select('string') | list }}"
```
sampaccoud, line 22: why do you need static-svc to coexist with svc? My understanding is that now all services will be removed from blue/green actions.

sampaccoud: sorry sorry I read your PR description after the code and now I understand. Smart :clap:

sampaccoud:
I'm afraid you need to copy the explanations in your PR description to a documentation somewhere in the project!

tasks/switch_static_services.yml
sampaccoud: Ok I get it, it's a prefix...

apps/edxapp/templates/services/nginx/route_cms.yml.j2
madmatah: I agree with you on the semantic, I used `prefix` because it's the current variable name on the master branch.

There is 60+ occurrences of this term in 25 files, I'm afraid it will complicate the review of this PR if I rename it here. Maybe I can address this in another PR, what do you think ?

sampaccoud: yes of course!

apps/edxapp/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,30 @@
...
+kind: Service
+metadata:
+  labels:
+    app: edxapp
+    service: nginx
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: Why not using a boolean here?
```suggestion
    removable: false
```

apps/edxec/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: edxec
+    service: nginx
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

apps/forum/templates/services/app/static-svc.yml.j2
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: forum
+    service: forum
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

apps/forum/templates/services/elasticsearch/static-svc.yml.j2
```diff
@@ -0,0 +1,24 @@
...
+kind: Service
+metadata:
+  labels:
+    app: forum
+    service: elasticsearch
+    version: "{{ forum_elasticsearch_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

apps/hello/templates/services/app/static-svc.yml.j2
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: hello
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 9: ditto

apps/learninglocker/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: learninglocker
+    service: nginx
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

apps/marsha/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: marsha
+    service: nginx
+    deployment_stamp: "{{ deployment_stamp }}"
+    service_prefix: "{{ prefix }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

apps/richie/templates/services/nginx/static-svc.yml.j2
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: richie
+    service: nginx
+    service_prefix: "{{ prefix }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+    type: static-service
+    removable: "no"
```
jmaupetit, line 10: ditto

tasks/delete_app_objects_kind.yml
jmaupetit, line 10:
```suggestion
      - removable=true
```

tasks/delete_app_objects_kind.yml
```diff
@@ -28,6 +29,7 @@
...
```
jmaupetit, line 33:
```suggestion
      - removable=true
```

apps/edxapp/templates/services/nginx/static-svc.yml.j2
madmatah: Kubernetes label values must be strings, so we can't use boolean values directly in yaml definitions. With your suggestion, it triggers the following error : 

> v1.Service.ObjectMeta: v1.ObjectMeta.Labels: ReadString: expects \\\\\" or n, but found f, error found in #10 byte of

As an alternative, we can use the following syntax : 

```yaml
removable: "false"
```

But I think this ambiguity in types increases the risk of errors, that's why I preferred to use "yes"/"no" values instead.

jmaupetit: Ok, then. Let's keep it as is.

madmatah:
> I'm afraid you need to copy the explanations in your PR description to a documentation somewhere in the project!

@sampaccoud 
I added basic documentation about supported kubernetes/openshift entities in `docs/developer_guide/concepts.md`.

I described static services and their purpose inside.

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+
+For the purpose of our blue/green deployment process, the name of a
+`DeploymentConfig` is dynamic: it is suffixed with a stamp that is unique to the version of the deployed application.
+
+### Service (aka `Blue/Green service`)
+
+[Service](https://kubernetes.io/docs/concepts/services-networking/service/)
+objects are declared in jinja2 template files matching the pattern `svc*.yml.j2`
+
+They are part of our blue/green our deployment process and their name are also dynamic: they are suffixed with a deployment stamp.
```
sampaccoud:
```suggestion
They are part of our blue/green deployment process and their name are also dynamic: they are suffixed with a deployment stamp.
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+Each deployed version of an application has its own set of services.
```
sampaccoud:
```suggestion
Each deployed version of an application has its own set of services and is referred to as a `stack`.
```
```suggestion
Each deployed version of an application has its own set of services.
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+Technically, they are pure kubernetes [Service](https://kubernetes.io/docs/concepts/services-networking/service/) objects.
+But in arnold, we use them for a specific case in our blue/green compatible applications.
+
+B/G Services are helpful for blue/green deployments and for internal communication inside the same application, but their dynamic name make it difficult to be used by other applications.
+That's why we introduced static services.
+
+Static services, as their name suggests, are services with a name that does not change.
+
+Static services are declined in 3 variants (i.e. the static service `foo` is declined in `foo-previous`, `foo-current` and `foo-next`).
+Each variant targets the corresponding version of the deployed application.
```
sampaccoud: sometimes you say version sometimes you say stack. See above comment.

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+
+The 2 use cases for a static service are:
+
+- Exposing an application service to another application
```
sampaccoud:
```suggestion
- Exposing an application to another application
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+- Exposing an application service to a route
```
sampaccoud:
```suggestion
- Exposing an application to a route
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+### Route
+
+[Route](https://docs.openshift.com/container-platform/3.11/architecture/networking/routes.html)
+objects are declared in jinja2 template files matching the pattern `route*.yml.j2`.
+
+> An OpenShift Container Platform route exposes a service at a host name, such as www.example.com, so that external clients can reach it by name.
+
+If you want to expose your application to external clients via HTTP, you must declare a route.
+
+Routes works with static services in a blue/green compatible application. So you must declare a static service if you declare a route.
```
sampaccoud:
```suggestion
Routes work with static services in a blue/green compatible application. So you must declare a static service if you declare a route.
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+
+Each route is declined in 3 variants:
+i.e. if you declare a route `bar`, it is declined in `bar-previous`, `bar` and `bar-next`.
```
sampaccoud: should we make it `bar-current` instead of `bar` for coherence with services?

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+
+And each variant targets the corresponding variant of the static service.
+
+### ConfigMap
+
+ConfigMap are automagically generated if you create files in the `configs` directory of a service template.
```
sampaccoud:
```suggestion
ConfigMaps are automagically generated if you create files in the `configs` directory of a service template.
```
The fact that you could use this word `automagically` speaks for a refactoring!

docs/developer_guide/concepts.md
madmatah: I think it's better now, please tell me if it's still confusing.

docs/developer_guide/concepts.md
madmatah: Oops it's a mistake, the route name is `bar-current` .
(I mixed it with the domain that is handled differently in the current route)

docs/developer_guide/concepts.md
madmatah: Yes it was a :wink:  to #253

madmatah:
I just added an `Upgrade.md` file to document the upgrade process.

Upgrade.md
```diff
@@ -0,0 +1,86 @@
...
+In order to migrate to this new system, you have to apply the following upgrade
+process. This will not cause any downtime, but it implies a redeployment of your
+applications.
+
+#### Upgrade Procedure
+
+This procedure must be applied for every arnold application deployed.
+
+The `current` stack of your application (before applying this procedure) will be
+referenced as `BEFORE-UPGRADE-VERSION`.
```
sampaccoud:
```suggestion
referred to as `BEFORE-UPGRADE-VERSION`.
```

Upgrade.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+##### 1. Upgrade your arnold version
+
+##### 2. Execute the `create_services` ansible playbook :
+
+```
+bin/ansible-playbook create_services.yml -e "apps_filter=your-application"
+```
+It will create the static  services required by your application. Each variant of these static services (`previous`, `current`, `next`) willtarget nothing after this step.
```
sampaccoud:
```suggestion
It will create the static services required by your application. Each variant of these static services (`previous`, `current`, `next`) will target nothing after this step.
```

Upgrade.md
```diff
@@ -0,0 +1,86 @@
...
+unchanged and available via the `current` route. On the other hand, the `next`
+stack you just deployed will not be reachable by any route.
+
+##### 4. Execute the `switch` ansible playbook :
+
+```
+bin/switch -e "apps_filter=your-application"
+```
+
+The switch will be done at a the static services level. So, the stack you just
```
sampaccoud:
```suggestion
The switch will be done at the static services level. So, the stack you just
```

Upgrade.md
```diff
@@ -0,0 +1,86 @@
...
+deployed will be considered as the `current` by static services and by the
+ansible playbooks.
+
+But again, this step does not affect routes. So your `BEFORE-UPGRADE-VERSION`
+stack stays unchanged and available via the `current` route.
+
+##### 5. Execute the `create_routes` playbook :
+
+```
+bin/ansible-playbook create_routes.yml -e "apps_filter=your-application"
```
sampaccoud, line 71: are you sure that this will not erase all SSL certificates?

madmatah: The routes already exist, so this playbook will only patch them with `deploy_patch_route.yml` to change the name of the targeted service, just like with the `switch_route.yml` task on master branch. So I assume nothing will be lost, right ?

sampaccoud: The last time I ran create_routes it completely reset all my routes. This has to be tested to be sure.

madmatah:
@lunika I patched `deploy_patch_cronjob.yml` because I had an error while testing the upgrade process after merging the cronjob feature.
Can you please review this part ?

Upgrade.md
madmatah: Yes, I can explain this behavior : when you execute the `create_routes` playbook of the master branch, it does not retrieve the deployment_stamp of the deployed stacks. As a result of what, the routes are patched to target an undefined service (i.e. route "hello-current" targets the service "hello-") and are broken.

The version of `create_routes` in this branch is simplified because it always points to a static service (i.e. route "hello-current" always targets service "hello-current").

I successfully tested the upgrade process on my local openshift cluster with the hello app.
The routes are patched and not re-created, so there should be no loss related to TLS certificates.

But of course we have to test it in a non-production environment.

madmatah:
I changed the base branch of this PR to target `migration-to-k8s`, the branch that we'll use to work on our Openshift to Kubernetes migration.

Maybe I should remove the Upgrade.md file because it's not relevant anymore.

apps/edxapp/templates/services/nginx/route_cms.yml.j2
sampaccoud: I opened an issue https://github.com/openfun/arnold/issues/601

apps/forum/templates/services/elasticsearch/static-svc.yml.j2
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: forum
+    service: elasticsearch
+    version: "{{ forum_elasticsearch_image_tag }}"
```
sampaccoud:
```suggestion
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
 
 * customize variables and files as necessary,
 * commit everything to the repository.
+
+## Supported object kinds
+
+Arnold applications can declare various kind of objects that will be deployed on the
```
sampaccoud:
```suggestion
Arnold applications can declare various kinds of objects that will be deployed on the
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+ensuring that the application stack is consistent.
+
+### Static Service
+
+*Static Service* objects are declared in jinja2 template files matching the pattern `static-svc*.yml.j2`.
+
+Technically, they are pure kubernetes [Service](https://kubernetes.io/docs/concepts/services-networking/service/) objects.
+But in arnold, we use them for a specific case in our blue/green compatible applications.
+
+B/G Services are helpful for blue/green deployments and for internal communication inside the same application, but their dynamic name make it difficult to be used by other applications.
```
sampaccoud:
```suggestion
B/G Services are helpful for blue/green deployments and for internal communication inside the same application, but their dynamic name makes it difficult to be used by other applications.
```

docs/developer_guide/concepts.md
```diff
@@ -159,3 +159,99 @@ After running this playbook, the developer will be invited to manually:
...
+To run a job after the deployment, you must set a label `job_type: "post"`
+
+### BuildConfig
+
+[BuildConfig](https://docs.openshift.com/container-platform/3.11/dev_guide/dev_tutorials/openshift_pipeline.html) are declared in jinja2 template files matching the pattern `bc_*.yml.j2`.
+
+### ImageStream
+
+[ImageStream](https://docs.openshift.com/container-platform/3.11/architecture/core_concepts/builds_and_image_streams.html) are declared in jinja2 template files matching the pattern `is_*.yml.j2`.
+
```
sampaccoud: Build configs and image streams were removed from Arnold as we now build our images in the different app repositories.
```suggestion
```

switch.yml
sampaccoud: the modification is still pending

tasks/deploy_patch_static_service.yml
```diff
@@ -0,0 +1,23 @@
+# Patch services of all services from a single app to the new deployment_stamp
```
sampaccoud:
```suggestion
# Patch all static services of a given app to the new deployment_stamp
```

tasks/deploy_patch_static_service.yml
```diff
@@ -0,0 +1,23 @@
...
+- name: "Patching services for the {{ app.name }} application"
+  k8s:
+    force: "{{ force_route | bool }}"
+    definition: "{{ lookup('template', service_template) | from_yaml }}"
+    state: present
+  loop: "{{ static_services }}"
+  loop_control:
+    loop_var: service_template
+  when: static_services | length > 0 and app.settings.is_blue_green_compatible | default(True) == False
+  tags: static_service
```
sampaccoud, line 23: These 2 steps look the same to me and I fail to understand why they can't be written in one. Is it just for logging purposes?

madmatah: Yes, I think it's for logging purpose only ! It's useful to track the progress during a switch.
it was the same thing on deploy_patch_route.yml before this PR.

sampaccoud: 👍

switch.yml
madmatah: Done :heavy_check_mark:

madmatah:
@sampaccoud Thanks for your feedback, I just pushed all your suggestions!

sampaccoud (approved):
Looks GTM. Just to make sure, mailcatcher and flower are not handled in this PR although they have routes. I suppose they have no B/G.

madmatah:
> Looks GTM. Just to make sure, mailcatcher and flower are not handled in this PR although they have routes. I suppose they have no B/G.

@sampaccoud   :tada: :pray: 

About `mailcatcher` and `flower` app, your assumptions are correct, you can see in their `vars/settings.yml` the following : 
```
is_blue_green_compatible: false
```

PR #429, opened 2020-01-09
:bug: (edxec) add edxec to eugene/development apps list

madmatah:
## Purpose

Can't deploy edxec on eugene/development because the app is unknown.

## Proposal

Add edxec too eugene/development apps list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
+- Add edxec to eugene/development apps list
+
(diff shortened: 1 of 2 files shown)


jmaupetit (approved):
The emoji in your commit title is not a unicode character but an emoji code.

madmatah:
> The emoji in your commit title is not a unicode character but an emoji code.

It's still the case now ?
In my first commit I put `:bug:` instead of the unicode character, but I fixed it and force pushed.

jmaupetit:
Looks like it's fixed! This is the PR title that misconducted me to this conclusion. Sorry for the noise.

:ship: it!

PR #432, opened 2020-01-11
✨(edxapp) create required directories in volumes and release Arnold 4.6.0

sampaccoud:
## Purpose

Some directories must pre-exist inside edxapp's volumes for the code to work. 

## Proposal

We propose to add a job that takes care of this on the basis of a variable listing all such required directories. The list of directories can thus be overriden for each project.


apps/edxapp/templates/services/cms/job_01_create_directories.yml.j2
```diff
@@ -0,0 +1,47 @@
...
+    spec:
+      containers:
+        - name: edxapp-cms-create-directories-{{ job_stamp }}
+          # We point to a local registry image build for this "live" image (see
+          # ImageStream and BuildConfig templates)
+          image: "{{ internal_docker_registry }}/{{ project_name }}/{{ image_name('cms', edxapp_image_tag, edxapp_theme_tag) }}"
+          command:
+            - "bash"
+            - "-c"
+            - {% for path in required_directories %}(mkdir {{ path }} || true) && {% endfor %}echo "✅"
```
jmaupetit: You should use `-p` in case of nested, deep paths and it always returns a success exit code.
```suggestion
            - {% for path in required_directories %}mkdir -p {{ path }} && {% endfor %}echo "✅"
```

apps/edxapp/templates/services/cms/job_01_create_directories.yml.j2
```diff
@@ -0,0 +1,47 @@
...
+              name: edxapp-v-media
+            - mountPath: /edx/app/edxapp/data
+              name: edxapp-v-data
+      volumes:
+        - name: edxapp-v-media
+          persistentVolumeClaim:
+            claimName: edxapp-pvc-media
+        - name: edxapp-v-data
+          persistentVolumeClaim:
+            claimName: edxapp-pvc-data
```
jmaupetit, line 58: I think we should mount all available volumes here?

apps/edxapp/templates/services/cms/job_01_create_directories.yml.j2
sampaccoud: :medal_sports:

apps/edxapp/templates/services/cms/job_01_create_directories.yml.j2
sampaccoud: you're right I mounted what I needed but since required directories= is configurable we should mount all.

apps/edxapp/vars/all/main.yml
```diff
@@ -72,6 +72,8 @@ edxapp_export_volume_size: 100Mi
 
 # -- volume names
 edxapp_export_volume_path: /edx/var/edxapp/export
+required_directories:
```
jmaupetit: Should be prefixed by the app name?
```suggestion
edxapp_required_directories:
```

sampaccoud: oh yes! So obvious I couldn't see it ^^

jmaupetit (approved):

PR #435, opened 2020-01-14
🐛(playbook) improve delete_previous playbook behaviour on k8s api fai…

madmatah:
## Purpose

The delete_previous playbook fails with an unclear error message if the kubernetes API is not responding :
```
fatal: [local]: FAILED! => {"msg": "The conditional check '( remaining_objects.resources | length ) == 0' failed. The error was: error while evaluating conditional (( remaining_objects.resources | length ) == 0): 'dict object' has no attribute 'resources'"}
```

## Proposal

Fix the retry loop condition in `delete_app_objects_kind.yml`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Avoid downtime during a switch or rollback
+- Improve `delete_previous` playbook behavior on k8s api failure
 
 ## [4.6.0] - 2020-01-13
 
--- a/tasks/delete_app_objects_kind.yml
+++ b/tasks/delete_app_objects_kind.yml
@@ -29,6 +29,6 @@
       - app={{ app.name }}
       - deployment_stamp={{ targeted_deployment_stamp }}
   register: remaining_objects
-  until: ( remaining_objects.resources | length ) == 0
+  until: remaining_objects.failed == True or ( remaining_objects.resources | length ) == 0
   retries: 120
   delay: 5


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Avoid downtime during a switch or rollback
+- Improve delete_previous playbook behaviour on k8s api failure
```
```suggestion
- Improve `delete_previous` playbook behavior on k8s api failure
```

tasks/delete_app_objects_kind.yml
```diff
@@ -29,6 +29,6 @@
       - app={{ app.name }}
       - deployment_stamp={{ targeted_deployment_stamp }}
   register: remaining_objects
-  until: ( remaining_objects.resources | length ) == 0
+  until: remaining_objects.failed == True or ( remaining_objects.resources | length ) == 0
```
jmaupetit, line 32: I am pretty sure this works, but I cannot find in Ansible's documentation a reference to this `failed` attribute for registered variables. Maybe this is not documented and you find it out by inspecting this variable upon failure... :man_shrugging:

madmatah: I found this in Ansible's documentation about [common return values](https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#failed).

CHANGELOG.md
madmatah: thanks

jmaupetit (approved):

tasks/delete_app_objects_kind.yml
jmaupetit, line 32: Something more pythonic to ensure `.failed` is a boolean and not the result of a comparison would be:

```suggestion
  until: remaining_objects.failed is True or ( remaining_objects.resources | length ) == 0
```

madmatah: I would like to be a better :snake: , but It gives me the following error : 

```
fatal: [local]: FAILED! => {"msg": "The conditional check 'remaining_objects.failed is True or ( remaining_objects.resources | length ) == 0' failed. The error was: template error while templating string: no test named 'True'. String: {% if remaining_objects.failed is True or ( remaining_objects.resources | length ) == 0 %} True {% else %} False {% endif %}"}
```

jmaupetit: Have you tried a lower case `true`?

madmatah: Yes, same error in lower case

sampaccoud: Maybe it is not a Python expression but Jinja?

jmaupetit: You are right: true is not a builtin test.
https://jinja.palletsprojects.com/en/2.10.x/templates/#list-of-builtin-tests

jmaupetit: Forget it.

Issue #416, opened 2019-12-31
Improve the switch playbook to avoid switching to nothing


sampaccoud:
## Bug Report

**Expected behavior/code**
If the `next` (blue) stack does not exist (nothing connected to the `next` route), the switch should not proceed with moving the `current` stack to `previous`... because the result is a downtime on the current route.

**Actual Behavior**
When running the `switch` playbook, the `current` stack will be moved to the `previous` route even if there is no replacing stack connected to the `next` route.

**Steps to Reproduce**
1. deploy a project
2. run `switch` twice

**Environment**
- Arnold version: 4.3.1

**Possible Solution**
Check that there is a stack connected to the `next` route before moving the `current` stack to the `previous` route.

PR #436, opened 2020-01-15
🚸(playbook) disallow switching if next stack is not deployed

madmatah:
## Purpose

As described in #416 : When running the `switch` playbook, the current stack will be moved to the previous  route even if there  is no replacing stack  connected to the next route. This can cause a service downtime.

## Proposal

Disallow switching if next stack is not deployed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -523,6 +523,47 @@ jobs:
             # Perform running pods test
             test $(eval "${cmd} | grep Running | wc -l") -eq 1
 
+  test-prevent-switch-to-nothing:
+    machine:
+      docker_layer_caching: false
+
+    working_directory: ~/fun
+
+    steps:
+      - checkout
+      - *attach_workspace
+      - *docker_load
+      - *ci_env
+      - *install_openshift_cluster
+      - *configure_openshift_cluster
+      - *run_openshift_cluster
+
+      - run:
+          name: Prepare "hello" application bootstrapping
+          command: |
+            # Bootstrap app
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Disallow switching if next stack is not deployed
```
```suggestion
- Prevent switching if next stack is not deployed
```

jmaupetit (commented):
Can you test it in the CI?

.circleci/config.yml
```diff
@@ -523,6 +523,47 @@ jobs:
...
+            # Test that the service is not responding on next route
+            bin/ci-test-service "hello" "Hello OpenShift! by Arnold" "/" "next" && exit 1 || true
+
+      - run:
+          name: Test that a second switch fails to execute with an error
+          command: |
+            # Execute the switch playbook and test that the exit code is not 0
+            bin/ci-ansible-playbook switch.yml "hello" &> /tmp/switch-to-nothing.out && exit 1 || true
+            # Test error message displayed by switch playbook
+            grep "next stack is not deployed" /tmp/switch-to-nothing.out
```
jmaupetit: Why not looking for the whole message?

```suggestion
            grep "next stack is not deployed, aborting switch" /tmp/switch-to-nothing.out
```

madmatah: idk

All checks have passed :tada: 
I'll commit your suggestion and I think we're good !

jmaupetit (approved):

Issue #112, opened 2018-08-30
Consider making statics volume persistant only for the stack lifetime


jmaupetit:
## Purpose

At the time of writing, static files are stored in a persistent volume (PVC) that is common to all stacks from an application for a particular customer / environment (OpenShift project). With a blue/green deployment strategy in mind, this may cause version conflicts when deploying a new release of our static files.

## Proposal

For each application using static files volumes, make those volumes unique per stack (_i.e._ by including the `deployment_stamp` in their name).

PR #445, opened 2020-01-22
[edxapp] switch to nginx-edxapp

jmaupetit:
## Purpose

Instead of serving static files from a volume where we have collected them, use an `nginx` image that bundles them.

## Proposal

- [x] use `edxapp-nginx` image
- [x] remove `edxapp`'s collectstatic jobs
- [x] remove `edxapp`'s static PV

Related to #112


apps/edxapp/templates/services/nginx/configs/cms.conf.j2
```diff
@@ -59,7 +59,7 @@ server {
     gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml image/svg+xml;
     expires {{ edxapp_nginx_cms_static_cache_expires }};
     add_header Cache-Control public;
-    root /data/static;
+    root /edx/app/edxapp/staticfiles;
```
jmaupetit, line 62: should add follow symlinks here

apps/edxapp/templates/services/nginx/configs/lms.conf.j2
```diff
@@ -86,7 +86,7 @@ server {
     gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml image/svg+xml;
     expires {{ edxapp_nginx_lms_static_cache_expires }};
     add_header Cache-Control public;
-    root /data/static;
+    root /edx/app/edxapp/staticfiles;
```
jmaupetit, line 89: ditto

apps/edxapp/templates/services/nginx/configs/cms.conf.j2
jmaupetit: According to the documentation, `nginx` follows symlinks by default: http://nginx.org/en/docs/http/ngx_http_core_module.html#disable_symlinks

sampaccoud (approved):

sampaccoud:
@jmaupetit ouch I just realized that we need a custom `edxapp-nginx` image for each customer because each theme comes with custom static files! Not sure if we can make both images with a bc in openshift :thinking:

jmaupetit:
@sampaccoud good point! I'll add an IS/BC for the edxapp/nginx service. I'll handle it in a new PR.

lunika (approved):

lunika:
IMO this change is a BC break

jmaupetit:
> IMO this change is a BC break

Yes. Should we write a migration doc?

lunika:
We can maybe add a `UPGRADE` file like we did on some other projects like Marsha ? https://github.com/openfun/marsha/blob/master/UPGRADE.md

PR #446, opened 2020-01-23
Bypass htaccess using a ip whitelist

lunika:
## Purpose

When a user has a trusted IP we wan't to deactivate the htaccess
protection while still keeping it active for other users.

## Proposal

Bypass htaccess using a whitelist of ip addresses for the following apps:

- [x] `richie`
- [x] `marsha`
- [x] `edxec`
- [x] `edxapp`

### DIFF
--- a/apps/edxapp/templates/services/nginx/configs/cms.conf.j2
+++ b/apps/edxapp/templates/services/nginx/configs/cms.conf.j2
@@ -1,3 +1,5 @@
+{% set http_basic_auth_enabled = activate_http_basic_auth or edxapp_activate_http_basic_auth %}
+{% set bypass_htaccess = edxapp_nginx_cms_bypass_htaccess_ip_whitelist | length > 0 %}
 upstream cms-backend {
   server edxapp-cms-{{ deployment_stamp }}:{{ edxapp_django_port }} fail_timeout=0;
 }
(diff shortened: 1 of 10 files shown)


apps/edxapp/templates/services/nginx/configs/cms.conf.j2
```diff
@@ -6,10 +7,10 @@ server {
   listen {{ edxapp_nginx_cms_port }};
   server_name localhost;
 
-  {% if activate_http_basic_auth or edxapp_activate_http_basic_auth -%}
+{% if http_basic_auth_enabled and edxapp_nginx_cms_bypass_htaccess_ip_whitelist | length == 0 %}
```
jmaupetit: Following the same pattern as for `http_basic_auth_enabled`, can you define a `bypass_htaccess` variable to improve readability?

apps/edxapp/templates/services/nginx/configs/cms.conf.j2
```diff
@@ -6,10 +7,10 @@ server {
...
   auth_basic "{{ http_basic_auth_message }}";
   auth_basic_user_file {{ http_basic_auth_user_file }};
-  {% endif %}
+{% endif %}
```
jmaupetit, line 22: The whole configuration is rather complex. I have the feeling that we can improve readability by nesting blocks as follow:

```jinja
{% if http_basic_auth_enabled %}
{% if bypass_htaccess %}
  location @basicauth {
    auth_basic "{{ http_basic_auth_message }}";
    auth_basic_user_file {{ http_basic_auth_user_file }};

    try_files $uri @proxy_to_cms_app;
  }
{% else %}
  auth_basic "{{ http_basic_auth_message }}";
  auth_basic_user_file {{ http_basic_auth_user_file }};
{% endif %}
{% endif %}
```

WDYT?

apps/edxapp/templates/services/nginx/configs/cms.conf.j2
lunika: :+1:

apps/edxapp/templates/services/nginx/configs/cms.conf.j2
lunika: :+1:

jmaupetit (approved):

PR #449, opened 2020-01-27
✨(deploy) manage cronjob object

lunika:
## Purpose

In some applications we would like to add cronjob objects. In order to
manage them we have a new regex mapping cronjob files (cronjob_*.yml.j2)
and then create them in the manage_app task. They are creating before
DeploymentConfig.

## Proposal

- [x] register cronjob templates in `get_objects_for_app` task.
- [x] create/delete cronjob object in `manage_app` task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   - `marsha`
   - `edxec`
   - `edxapp`
+- Handle `CronJob` object type
 
 ### Changed
 
--- a/apps/hello/templates/services/app/cronjob_hello.yml.j2
+++ b/apps/hello/templates/services/app/cronjob_hello.yml.j2
@@ -1,0 +1,33 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: "cronjob-hello-{{ deployment_stamp }}"
+  namespace: "{{ project_name }}"
+  labels:
+    app: hello
+    service: app
+    deployment_stamp: "{{ deployment_stamp }}"
+spec:
+  schedule: "*/3 * * * *"  
+  successfulJobsHistoryLimit: 2
+  failedJobsHistoryLimit: 1
+  concurrencyPolicy: Forbid
+  suspend: {{ suspend_cronjob | default(false) }}
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          name: "cronjob-hello-{{ deployment_stamp }}"
+          labels:
+            app: hello
+            service: app
+            deployment_stamp: "{{ deployment_stamp }}"
+        spec:
+          containers:
+            - name: hello-world
+              image: busybox:1.31.1
+              command:
+                - "sh"
+                - "-c"
+                - echo "Hello World!"
+          restartPolicy: Never
--- a/tasks/delete_app.yml
+++ b/tasks/delete_app.yml
@@ -6,15 +6,16 @@
 - name: Iterate over objects to delete with deployment_stamp {{ targeted_deployment_stamp }} for app {{ app.name }}
   include_tasks: tasks/delete_app_objects_kind.yml
   loop:
-    - DeploymentConfig
+    - { kind: "DeploymentConfig", api_version: "v1" }
     # Manually handle deletion cascade as the k8s module won't
     # https://github.com/ansible/ansible/issues/42302
-    - ReplicationController
-    - Pod
-    - Job
-    - Service
-    - Endpoints
-    - ConfigMap
+    - { kind:  "ReplicationController", api_version: "v1" }
+    - { kind:  "Pod", api_version: "v1" }
+    - { kind:  "Job", api_version: "v1" }
+    - { kind:  "CronJob", api_version: "v1beta1" }
+    - { kind:  "Service", api_version: "v1" }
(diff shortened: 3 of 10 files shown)


madmatah (commented):
Do you plan to handle the `switch` problem (to avoid cronjob from multiple stacks to be active simultaneously) in another PR ? I've found that we can use `spec.suspend: true` in a cronjob definition to prevent new jobs from being run.

lunika:
> Do you plan to handle the `switch` problem (to avoid cronjob from multiple stacks to be active simultaneously) in another PR ? I've found that we can use `spec.suspend: true` in a cronjob definition to prevent new jobs from being run.

I didn't start working on the switch part, but with what you found I will work on this PR. Thanks

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   - `marsha`
   - `edxec`
   - `edxapp`
+- Manage cronjob objects
```
```suggestion
- Handle `CronJob` object type
```

lunika:
@madmatah switch and delete are handle now

madmatah (commented):
Can you please provide an example of cronjob template definition in `hello` app ?

It will be useful as an example and it will allow CircleCI to test your ansible tasks.

tasks/switch_cronjobs.yml
```diff
@@ -0,0 +1,26 @@
+---
+# In a switch, cronjobs in previous stacks must be suspended.
```
jmaupetit:
```suggestion
# In a switch, cronjobs from the previous stack must be suspended.
```

tasks/switch_cronjobs.yml
```diff
@@ -0,0 +1,26 @@
...
+# CronJobs in the current stack must be activated
```
jmaupetit:
```suggestion
# and those from the current stack must be activated.
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
   - `marsha`
```
sampaccoud, line 20: In the commit message : 
s/They are creating before DeploymentConfig/They are created before DeploymentConfig

tasks/delete_app.yml
```diff
@@ -6,15 +6,16 @@
 - name: Iterate over objects to delete with deployment_stamp {{ targeted_deployment_stamp }} for app {{ app.name }}
```
sampaccoud, line 6: in commit message:
s/shlould/should
s/wah/was
s/ans use/and used
s/delete_app_objects_kinf/delete_app_objects_kind

tasks/delete_app_objects_kind.yml
```diff
@@ -1,30 +1,31 @@
 # Delete a kind of objects for the targeted deployment_stamp
-- name: Lookup {{ kind }} to delete with deployment_stamp {{ targeted_deployment_stamp }} for app {{ app.name }}
+- name: Lookup {{ resource_to_delete.kind }} to delete with deployment_stamp {{ targeted_deployment_stamp }} for app {{ app.name }}
```
sampaccoud, line 2: Why don't you remove the `version` selector so that we delete all objects of this kind whatever their version?

apps/hello/templates/services/app/cronjob_hello.yml.j2
```diff
@@ -0,0 +1,33 @@
+apiVersion: batch/v1beta1
```
sampaccoud, line 1: in commit message:
s/currect/correct
s/the other one/the other ones
s/active only if the are switc in the current task/activated only when they are switched to the current stack

tasks/delete_app_objects_kind.yml
lunika: because by default the value is `v1` see https://docs.ansible.com/ansible/latest/plugins/lookup/k8s.html

sampaccoud: wow that sucks. Have you tried setting it to null?

lunika: No, do you think it's better than what I made ?

sampaccoud: well it would be simpler and would avoid looping on all versions when you get several?

lunika: I don't think we loop on all versions, we use only the one defined in the hashmap

tasks/delete_app.yml
jmaupetit: :joy:

lunika: I didn't laugh...

sampaccoud: @jmaupetit so you're the referee on four legs?

tasks/delete_app_objects_kind.yml
sampaccoud: Ok up to you but it is also uglier that you have to pass a whole object to define what you want to delete... so if the version could be ignored it would be better imo. If null doesn't work then ok lah.

lunika: I tried and no resources are deleted, the `delete_app` fails with a timeout

sampaccoud (approved):

madmatah (approved):

Issue #450, opened 2020-01-28
Unable to put a valid JSON file to a ConfigMap (from a Jinja2 template)


madmatah:
## Bug Report

**Expected behavior/code**

I would like to generate a JSON file from a Jinja2 template, in order to put it in a ConfigMap.

**Actual Behavior**

The deploy playbook fail with the following error in the task _Create the config maps in OpenShift_ : 

```
failed: [local] (item={'key': 'app', 'value': {'test.json': {'stamp': 'd-200128-08h49m48s', 'project': 'development-eugene'}}}) => {"ansible_loop_var": "item", "changed": false, "error": 400, "item": {"key": "app", "value": {"test.json": {"project": "development-eugene", "stamp": "d-200128-08h49m48s"}}}, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"ConfigMap in version \\\\\"v1\\\\\" cannot be handled as a ConfigMap: [pos 242]: json: expect char \\'\\\\\"\\' but got char \\'{\\'\",\"reason\":\"BadRequest\",\"code\":400}\\n'", "reason": "Bad Request", "status": 400}
```

**Steps to Reproduce**
1. Create a file `apps/hello/templates/services/app/configs/test.json.j2` with the following content : 
```json
{
  "stamp": "{{ deployment_stamp }}",
  "project": "{{ project_name }}"
}
```
2. Deploy the hello app
3. And then the bug happens!

**Environment**
- Arnold version: 4.7.0 on `master`
- Platform: linux / oc 3.11

**Possible Solution**

I tried to patch the _Format configMaps_ task in `tasks/create_app_config.yml` with an explicit cast to string for every rendered templates : 

```
{%- set content = lookup('template', cm_path)|string -%}
```

The `deploy` playbook is not failing anymore, but if I inspect the content of the generated ConfigMap, I can see that my JSON content has been altered (double quotes are converted to single quotes) and is invalid : 

```
$> oc describe cm hello-app-d-200128-09h01m37s
Name:         hello-app-d-200128-09h01m37s
Namespace:    development-eugene
Labels:       app=hello
              deployment_stamp=d-200128-09h01m37s
              service=app
Annotations:  <none>

Data
====
test.json:
----
{'stamp': 'd-200128-09h01m37s', 'project': 'development-eugene'}
Events:  <none>
```

**Additional context/Screenshots**

[...]

PR #451, opened 2020-01-28
🐛(playbook) fix JSON config file generation from jinja2 template

madmatah:
## Purpose

As described in #450, the deploy playbook fails to generate a ConfigMap if it contains a JSON file generated from a Jinja2 template.

## Proposal

Prevent ansible from converting/interpreting files that will be stored in ConfigMaps.

**Do you see any side effects generated by this proposal ?**

### DIFF
--- a/apps/hello/templates/services/app/configs/test.json.j2
+++ b/apps/hello/templates/services/app/configs/test.json.j2
@@ -1,0 +1,5 @@
+{
+  "stamp": "{{ deployment_stamp }}",
+  "project": "{{ project_name }}",
+  "key": "valu€"
+}
--- a/tasks/create_app_config.yml
+++ b/tasks/create_app_config.yml
@@ -36,7 +36,7 @@
       {%- for service in app_configmaps -%}
         {%- set _ = configmaps.update({ service.name: {} }) -%}
         {%- for cm_path in service.configs -%}
-          {%- set content = lookup('template', cm_path) -%}
+          {%- set content = lookup('template', cm_path, convert_data = False) -%}
           {%- set filename = cm_path | basename | regex_replace('.j2') -%}
           {%- set _ = configmaps.get(service.name).update({filename: content}) -%}
         {%- endfor -%}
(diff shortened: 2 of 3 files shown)


apps/hello/templates/services/app/configs/test.json.j2
```diff
@@ -0,0 +1,5 @@
+{
+  "stamp": "{{ deployment_stamp }}",
+  "project": "{{ project_name }}",
+  "key": "valu€"
+}
```
jmaupetit: A carriage return is missing here.

apps/hello/templates/services/app/configs/test.json.j2
```diff
@@ -0,0 +1,5 @@
+{
```
jmaupetit, line 1: Have you tried to mount this configMap in the hello app DC?

tasks/create_app_config.yml
```diff
@@ -36,7 +36,7 @@
       {%- for service in app_configmaps -%}
         {%- set _ = configmaps.update({ service.name: {} }) -%}
         {%- for cm_path in service.configs -%}
-          {%- set content = lookup('template', cm_path) -%}
+          {%- set content = lookup('template', cm_path, convert_data = False) -%}
```
jmaupetit, line 39: Looks like a smart fix :1st_place_medal:

apps/hello/templates/services/app/configs/test.json.j2
madmatah: I didn't try to mount the configMap because if the object is deployed on kubernetes without any error, I assume that the configMap is mountable.

I checked its content manually and I added the `test.json.j2` file to the project to ensure that the playbook doesn't fail on the CI.

To have a complete test suite, I can test the validity of the generated json file by mounting it on a container with `jq` utility installed, but do you think it's worth it ?

tasks/create_app_config.yml
madmatah: Thanks !

apps/hello/templates/services/app/configs/test.json.j2
jmaupetit: Nope. Go on!

jmaupetit (approved):
Once you've fixed this missing carriage return, merge this!

lunika (approved):

PR #456, opened 2020-02-05
Fixes for elasticsearch app

madmatah:
## Purpose

3 issues fixed in this PR :

1. Too many false negative in circleCI
2. The `elasticsearch_memory_lock` variable is always considered true by elasticsearch
3. Unable to discover peers on some environments because of DNS caching

## Proposal

1. Remove `test-bootstrap-elasticsearch` test from CircleCI
2. Use lowercase boolean values in `elasticsearch.yml`
3. Use a classic `ClusterIP` kubernetes service instead of a headless service

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - `elasticsearch-discovery` service is not a headless service anymore
+- Generate a valid YAML value from `elasticsearch_memory_lock` variable
 
 
 ## [5.0.0] - 2020-01-31
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
+- `elasticsearch-discovery` service is not a headless service anymore
+
+
 ## [5.0.0] - 2020-01-31
 
 ### Added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
+- `elasticsearch-discovery` service is not a headless service anymore
+- the `elasticsearch_memory_lock` variable was always considered true
```
lunika: Maybe changing to:
```suggestion
- set a valid json value in `elasticsearch_memory_lock` variable
```

WDYT ?

madmatah: In fact, the configuration file in which this value is written is a YAML file (`elasticsearch.yml`).

According to the [YAML spec](https://yaml.org/type/bool.html), `False` should be a correct boolean value. But elasticsearch thinks it's `true`.

I used `to_json` to ensure that the value is lowercase, but I think I should have used `to_yaml` instead. The result is the same, but it will be less confusing.

So I'll follow your suggestion but write `yaml` instead of `json`.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - `elasticsearch-discovery` service is not a headless service anymore
+- generate a valid yaml value from `elasticsearch_memory_lock` variable
```
```suggestion
- Generate a valid YAML value from `elasticsearch_memory_lock` variable
```

jmaupetit (approved):

PR #457, opened 2020-02-06
✅(scripts) add argument to ci-test-route to set the max retries

madmatah:
## Purpose

The `ci-test-route`  script has  a hardcoded value  (5) for  the maximum number of retries to get the  expected response from the server before failing. 

This  number  is   too  low  for  the  `test-bootstrap-edxapp` test suite and  generates a lot  of false negative  in the CI.  

## Proposal

- Add a new argument  to the  script to  specify the  maximum number  of retries.

- Increase max_retry to 20 for `test-bootstrap-edxapp` tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Add argument to `ci-test-route` to configure the maximum number of retries
+
 ### Fixed
 
 - `elasticsearch-discovery` service is not a headless service anymore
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Add argument to `ci-test-route` to configure the maximum number of retries
+
```
jmaupetit, line 14: As it is related to the project, I'm not sure we should log this... WDYT @lunika?

jmaupetit (approved):

CHANGELOG.md
lunika: I think we can keep this new entry in the changelog.

madmatah: Thanks ! I'll merge this one when the CI checks are finished then.

PR #462, opened 2020-02-27
🐛(edxapp) Copy staticfiles in nginx_collectstatic build config

lunika:
## Purpose

Collecting static files where done only if a theme is used. The build
`bc_nginx` where failing because the directory /edx/app/edxapp/staticfiles
does not exists if a theme is not used. We should always copy
staticfiles in the collectstatic build config.

## Proposal

- [x] always copy staticfiles in the collectstatic build config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - `elasticsearch-discovery` service is not a headless service anymore
 - Generate a valid YAML value from `elasticsearch_memory_lock` variable
 - Set `edxapp`'s jobs image pull policy to "Always"
+- Copy staticfiles in `edxapp` `nginx_collectstatic` build config
 
 ## [5.0.0] - 2020-01-31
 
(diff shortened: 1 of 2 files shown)


jmaupetit (approved):
My bad... sorry. :pray:

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - `elasticsearch-discovery` service is not a headless service anymore
```
jmaupetit, line 21: In your commit message:

> Static files were collected only when a custom theme was configured, leading, the `nginx` BC to fail since the  `/edx/app/edxapp/staticfiles` directory does not exist. Static files should always be collected whether a theme has been configured or not.

lunika:
@jmaupetit I needed to update the PR. Ruby was also missing in the collecstatic image. You should make a new review please

lunika:
This fix is not the good one to implement. I will create a new PR with a better implementation.

PR #470, opened 2020-03-18
✨(ashley) add app Ashley, the alternative discussion forum for OpenEdx

madmatah:
## Purpose

We want to deploy our new projet [ashley](https://github.com/openfun/ashley) with Arnold.

Ashley is an alternative discussion forum for OpenEdx based on LTI.

## Proposal

I basically copied the richie app in Arnold to adapt it to ashley.


.circleci/config.yml
```diff
@@ -478,6 +478,41 @@ jobs:
             # Test service switched to the current route
             bin/ci-test-route "learninglocker" "OK" "/api"
 
+
+  # Test the bootstrap playbook on the "richie" application
```
jmaupetit:
```suggestion
  # Test the bootstrap playbook on the "ashley" application
```

apps/ashley/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,100 @@
...
+            - name: DB_ENGINE
+              value: "{{ ashley_database_engine }}"
+            - name: DB_NAME
+              value: "{{ ashley_database_name }}"
+            - name: DB_HOST
+              value: "ashley-{{ ashley_database_host }}-{{ deployment_stamp }}"
+            - name: DB_PORT
+              value: "{{ ashley_database_port }}"
+            - name: DJANGO_ALLOWED_HOSTS
+              value: "{{ ashley_host | blue_green_hosts }}"
```
jmaupetit: All those definitions should stand in the dot env configMap?

apps/ashley/templates/services/app/job_01_collectstatic.yml.j2
```diff
@@ -0,0 +1,62 @@
...
+            - name: DJANGO_CONFIGURATION
+              value: "{{ ashley_django_configuration }}"
+            - name: DB_ENGINE
+              value: "{{ ashley_database_engine }}"
+            - name: DB_NAME
+              value: "{{ ashley_database_name }}"
+            - name: DB_HOST
+              value: "ashley-{{ ashley_database_host }}-{{ deployment_stamp }}"
+            - name: DB_PORT
+              value: "{{ ashley_database_port }}"
```
jmaupetit: ditto 

And you should also load environment variables from the dot env configMap.

apps/ashley/templates/services/app/job_02_db_migrate.yml.j2
```diff
@@ -0,0 +1,51 @@
...
+            - name: DJANGO_CONFIGURATION
+              value: "{{ ashley_django_configuration }}"
+            - name: DB_ENGINE
+              value: "{{ ashley_database_engine }}"
+            - name: DB_NAME
+              value: "{{ ashley_database_name }}"
+            - name: DB_HOST
+              value: "ashley-{{ ashley_database_host }}-{{ deployment_stamp }}"
+            - name: DB_PORT
+              value: "{{ ashley_database_port }}"
```
jmaupetit: Ditto.

apps/ashley/templates/services/app/secret.yml.j2
```diff
@@ -0,0 +1,15 @@
...
+    app: ashley
+    service: app
+  name: "{{ ashley_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  DJANGO_SECRET_KEY: "{{ ASHLEY_VAULT.DJANGO_SECRET_KEY | default('supersecret') | b64encode }}"
+{% if ASHLEY_VAULT.DJANGO_SENTRY_DSN is defined and ASHLEY_VAULT.DJANGO_SENTRY_DSN is not none %}
+  DJANGO_SENTRY_DSN: "{{ ASHLEY_VAULT.DJANGO_SENTRY_DSN | b64encode }}"
+{% endif %}
+  DB_USER: "{{ ASHLEY_VAULT.DB_USER | default('edxapp_user') | b64encode }}"
```
jmaupetit, line 14:
```suggestion
  DB_USER: "{{ ASHLEY_VAULT.DB_USER | default('ashley') | b64encode }}"
```

apps/ashley/vars/settings.yml
```diff
@@ -0,0 +1,3 @@
+databases:
+  - engine: "postgresql"
+    release: "12"
```
jmaupetit: Sure? Will it work with potsgresql 9.6?

sampaccoud: Our production is running posgresql 12.

jmaupetit: Yes, but this is a (unnecessary?) strong requirement and not consistent with other apps like [richie](https://github.com/openfun/arnold/blob/master/apps/richie/vars/settings.yml#L3) or [marsha](https://github.com/openfun/arnold/blob/master/apps/marsha/vars/settings.yml#L3).

sampaccoud: Ah ok. I misunderstood your comment.

madmatah: Thanks for your feedbacks :slightly_smiling_face: 

:man_shrugging: I do not realize all the consequences that this implies to set version 12 here. It defines a requirement, or just a variable that one can override if he wants to use another version ?

I use postgresql 12 in `ashley` since [this suggestion](https://github.com/openfun/ashley/pull/3#discussion_r385197418), so I put it everywhere to be consistent. Do I have to be consistent with other projects, or ashley ? :slightly_smiling_face:

jmaupetit: In case someone else wants to deploy it at home, we need to ensure the largest compatibility with database engines.

madmatah: I've set version `9.6`, but I'm not sure this is the solution to ensure the largest compatibility with database engines. Now we don't know if it works on latest releases.

To achieve this, another approach would be to run the CI on multiple database versions in ashley project, but it's out of the scope of this PR.

sampaccoud:
@madmatah this is more easily done in Ashley's repo...?

madmatah:
> @madmatah this is more easily done in Ashley's repo...?

yes

apps/ashley/templates/services/nginx/configs/healthcheck.conf.j2
```diff
@@ -0,0 +1,9 @@
+server {
+  listen {{ ashley_nginx_healthcheck_port }};
+  server_name localhost;
+
+  location {{ ashley_nginx_healthcheck_endpoint }} {
+    add_header Content-Type text/plain;
```
jmaupetit, line 6: I've added the following for `etherpad` to not pollute logs, WDYT?

```suggestion
    add_header Content-Type text/plain;
    access_log off;
```

If you agree, this should be backported for other apps IMO (in a new PR).

jmaupetit (approved):

apps/ashley/templates/services/nginx/configs/healthcheck.conf.j2
madmatah: I agree with you,  good suggestion ! :+1:

PR #473, opened 2020-03-23
Add etherpad application

jmaupetit:
## Purpose

We need to deploy [etherpad](https://etherpad.org) instances that can be further integrated in our services.

## Proposal

- [x] add base application
- [x] test application in the CI


apps/etherpad/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,67 @@
...
+            - node node_modules/ep_etherpad-lite/node/server.js --sessionkey private/SESSIONKEY.txt --apikey private/APIKEY.txt
+          livenessProbe:
+            httpGet:
+              path: /
+              port: {{ etherpad_application_port }}
+            initialDelaySeconds: 30
+            periodSeconds: 10
+          readinessProbe:
+            httpGet:
+              path: /
```
lunika, line 54: You are just testing is the application is up and running ? Should we do that in the probes without knowing why the application is failing  ?

apps/etherpad/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,67 @@
...
+            initialDelaySeconds: 5
+            periodSeconds: 5
+          envFrom:
+            - secretRef:
+                name: "{{ etherpad_env_secret_name }}"
+            - configMapRef:
+                name: "etherpad-app-dotenv-{{ deployment_stamp }}"
+          volumeMounts:
+            - name: etherpad-v-secret
+              mountPath: /app/private
```
lunika: should we use a variable for this path ? We are using it in different files, using a variable will prevent errors if this path changes

apps/etherpad/templates/services/app/dc.yml.j2
jmaupetit: I have no better idea at the time without modifying the application itself. Maybe we should consider integrating [terminus](https://github.com/godaddy/terminus) or [lightship](https://github.com/gajus/lightship) to add health check views to the express server but I am not sure if this can be achieved _via_ plugins development or just server script hacking.

jmaupetit: It looks like there is a server hook for that: https://etherpad.org/doc/v1.6.2/#index_expresscreateserver but we must create the plugin that integrates lightship.

jmaupetit: As this require some work to add a plugin to the image, if you agree, I would like to fix this in a separate PR. WDYT?

apps/etherpad/vars/vault/main.yml.j2
```diff
@@ -0,0 +1,11 @@
+# customer: {{ customer }}
+# env_type: {{ env_type }}
+
+# postgresql
+{% set postgresql_credentials = databases.postgresql | json_query("[?release=='" ~ etherpad_postgresql_version ~ "'].databases | [0][?application=='etherpad'].{user: user, password: password} | [0]") %}
```
sampaccoud, line 5: why not separate the variables to make it simpler?

apps/etherpad/templates/services/app/dc.yml.j2
sampaccoud: So open an issue?

jmaupetit: I was waiting for @lunika answer :sweat_smile: Done, see #489

apps/etherpad/vars/vault/main.yml.j2
jmaupetit: If I you mean separating `user` from `password`, by doing so, you will duplicate the same request with the same complexity. Note that this request is used for almost every app vault file.

apps/etherpad/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,77 @@
...
+        - name: etherpad
+          image: "{{ etherpad_image_name }}:{{ etherpad_image_tag }}"
+          imagePullPolicy: Always
+          command:
+            - "bash"
+            - "-c"
+            - node node_modules/ep_etherpad-lite/node/server.js
+                --settings {{ etherpad_configs_directory_path }}/settings.json
+                --sessionkey {{ etherpad_private_directory_path }}/SESSIONKEY.txt
+                --apikey {{ etherpad_private_directory_path }}/APIKEY.txt
```
madmatah: You should add double quotes around these parameters to avoid problems with directories containing spaces.

apps/etherpad/templates/services/app/secret_keys.yml.j2
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: etherpad
+    service: etherpad
+  name: "{{ etherpad_file_secret_name }}"
+  namespace: "{{ project_name }}"
+data:
+  SESSIONKEY.txt: "{{ ETHERPAD_VAULT.SESSIONKEY | default('thisisasessionkey') | b64encode }}"
```
madmatah: Should we fail instead of defining a default session key ?

apps/etherpad/templates/services/nginx/configs/healthcheck.conf.j2
```diff
@@ -0,0 +1,10 @@
+server {
+  listen {{ etherpad_nginx_healthcheck_port }};
+  server_name localhost;
+
+  location {{ etherpad_nginx_healthcheck_endpoint }} {
+    add_header Content-Type text/plain;
+    access_log off;
+    return 200 "OK";
+  }
+}
```
madmatah, line 14: You should enable the nginx status page (see #479 )

apps/etherpad/templates/services/app/dc.yml.j2
jmaupetit: Please, don't tell me people are doing this?

madmatah: It's not forbidden !

apps/etherpad/templates/services/app/secret_keys.yml.j2
jmaupetit: @sampaccoud asked me a similar question. IMO this should be a global policy for all vault-stored values.

jmaupetit: But I can change this here if you insist a bit :sweat_smile:

madmatah: If you change this here, it would show us the way to a better future :wink:

apps/etherpad/templates/services/nginx/configs/healthcheck.conf.j2
jmaupetit: Good catch!

apps/etherpad/templates/services/app/dc.yml.j2
jmaupetit: Done.

apps/etherpad/templates/services/app/secret_keys.yml.j2
jmaupetit: Well said dude!

jmaupetit:
@madmatah I've added fixup commits so that it's easier to give it a second review :pray:

madmatah (approved):

PR #475, opened 2020-03-26
✨(core) upgrade acme to v2 API

jmaupetit:
## Purpose

OpenShift-acme project recently added the Let's encrypt API v2 support. Switching to this version of the API is required since previous endpoint were considered as deprecated and closed.

## Proposal

- [x] integrate latest `openshift-acme` objects [1]

[1] https://github.com/tnozicka/openshift-acme/tree/master/deploy/single-namespace


core_apps/acme/templates/services/app/cm_live.yml.j2
```diff
@@ -0,0 +1,11 @@
+kind: ConfigMap
```
lunika, line 1: I don't understand how these config maps are used ?

madmatah: If I understand correctly... the `openshift-acme` controller get these ConfigMap by matching their labels :

 - https://github.com/tnozicka/openshift-acme/blob/8b54945f9457b1b960fea133bce359679eee5d62/pkg/api/internal.go#L22-L25
- https://github.com/tnozicka/openshift-acme/blob/8b54945f9457b1b960fea133bce359679eee5d62/pkg/controllerutils/controllerutils.go#L35

And they define the available certificate issuers.

madmatah (commented):
In the [single-namespace deployment example](https://github.com/tnozicka/openshift-acme/tree/master/deploy/single-namespace), a [service account](https://github.com/tnozicka/openshift-acme/blob/master/deploy/single-namespace/serviceaccount.yaml) is defined. Why isn't it required here ?

madmatah:
> In the [single-namespace deployment example](https://github.com/tnozicka/openshift-acme/tree/master/deploy/single-namespace), a [service account](https://github.com/tnozicka/openshift-acme/blob/master/deploy/single-namespace/serviceaccount.yaml) is defined. Why isn't it required here ?

Oh sorry I did not see the `sa.yml` file... :man_facepalming:

madmatah (approved):

core_apps/acme/templates/services/app/cm_live.yml.j2
jmaupetit: You'll have to checkout the `openshift-acme` code to understand this part.

jmaupetit: (I didn't see @madmatah comment before answering)

PR #477, opened 2020-03-30
Edxapp fix static files url computation

sampaccoud:
## Purpose

Arnold v5.0.0 introduced an `edxapp-nginx` docker image that includes static files for performance improvement. This feature was broken because the `edxapp` container could not access static files anymore and it is required by the `CachedStaticFilesStorage` backend to calculate file hashes.

## Proposal

Bringing back the static volume allows running the `collectstatic` command manually in `edxapp` containers so that `Django` can access the files at run time. This buys us time to work on a real fix.

I'm also proposing to stop declaring a specific tag for the `edxapp-nginx` docker image. Instead, we can use the `edxapp` image tag for both the `edxapp` and `nginx` images to make sure they are always in sync.


CHANGELOG.md
```diff
@@ -17,6 +17,16 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 - Upgrade PyYAML to `5.3.1`
 - Upgrade `acme` core app to its latest release that supports ACME v2 API
 
+### Removed
+
+- Stop declaring a specific tag for the `edxapp-nginx` docker image and use
+  the `edxapp` image tag for both the `edxapp` and `nginx` images
```
jmaupetit, line 23: Good idea!

CHANGELOG.md
```diff
@@ -17,6 +17,16 @@ Versioning](http://semver.org/spec/v2.0.0.html).
...
+
+### Fixed
+
+- Temporarily reinstate the static volume on edxapp pods to allow using the
+  `nginx` docker image with bundled static files, while waiting for a fix
```
jmaupetit: WDYT?
```suggestion
  `nginx` docker image with bundled static files, while waiting for a long-term fix
```

jmaupetit (approved):

CHANGELOG.md
sampaccoud: Good idea!

PR #481, opened 2020-04-03
🔧(nextcloud) remove app skeleton in BuildConfig

lunika:
## Purpose

The app contains a skeleton used when a new user is created. This
skeleton is used to create demonstration files and we don't want them.
Moreover removing them increase the performace when a user is created
from SAML provider.

## Proposal

- [x] remove skeleton in the BuildConfig


sampaccoud (approved):
:man_shrugging:

jmaupetit (commented):
You forgot to update the changelog :pray:

lunika:
> You forgot to update the changelog pray

There is no need, we didn't release Arnold shipping nextcloud.

PR #486, opened 2020-04-07
✨(nextcloud) install user_saml app in BuildConfig

lunika:
## Purpose

We want to install user_saml application in the BuildConfig. The version
to install is configurable with nextcloud_apps_user_saml_version
variable.

## Proposal

- [x] Install user_saml app in nextcloud BuildConfig


lunika:
This work should be done in matsuo and not in Arnold. I close this PR

PR #495, opened 2020-04-20
⚡️(etherpad) implement reliable health checks

### LINKED ISSUE #489: [etherpad] improve health checks once implemented in the docker image
## Purpose

Current `etherpad` app health checks are very low level checking (waiting for a code 200 HTTP response). We need to work on this to ensure that the application is up and running.

## Proposal

Once [lightship](https://github.com/gajus/lightship) has been integrated to our [stateless etherpad docker image](https://github.com/openfun/etherpad-docker), we need to update `etherpad`'s `app` container probes.

jmaupetit:
## Purpose

Now that lightship has been integrated in our Docker image [1], liveness and readyness probes could be implemented consequently.

[1] https://github.com/openfun/etherpad-docker/commit/697b2e368b1cd9ea6921e2ea9c4140d481cd1e80

## Proposal

- [x] use lightship endpoints as health check probes

Fix #489

### DIFF
--- a/apps/etherpad/templates/services/app/dc.yml.j2
+++ b/apps/etherpad/templates/services/app/dc.yml.j2
@@ -45,16 +45,26 @@ spec:
                 --apikey "{{ etherpad_private_directory_path }}/APIKEY.txt"
           livenessProbe:
             httpGet:
-              path: /
-              port: {{ etherpad_application_port }}
-            initialDelaySeconds: 30
-            periodSeconds: 10
+              path: {{ etherpad_app_healthcheck_live_endpoint }}
+              port: {{ etherpad_app_healthcheck_port }}
+            failureThreshold: 3
+            initialDelaySeconds: 10
+            # Allow sufficient amount of time (90 seconds = periodSeconds * failureThreshold)
+            # for the registered shutdown handlers to run to completion.
+            periodSeconds: 30
+            successThreshold: 1
+            # Setting a very low timeout value (e.g. 1 second) can cause false-positive
+            # checks and service interruption.
+            timeoutSeconds: 5
           readinessProbe:
             httpGet:
-              path: /
-              port: {{ etherpad_application_port }}
+              path: {{ etherpad_app_healthcheck_ready_endpoint }}
+              port: {{ etherpad_app_healthcheck_port }}
+            failureThreshold: 3
             initialDelaySeconds: 5
             periodSeconds: 5
+            successThreshold: 1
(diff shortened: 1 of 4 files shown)


apps/etherpad/templates/services/app/dc.yml.j2
```diff
@@ -45,16 +45,26 @@ spec:
...
+            # Setting a very low timeout value (e.g. 1 second) can cause false-positive
+            # checks and service interruption.
+            timeoutSeconds: 5
           readinessProbe:
             httpGet:
-              path: /
-              port: {{ etherpad_application_port }}
+              path: {{ etherpad_app_healthcheck_ready_endpoint }}
+              port: {{ etherpad_app_healthcheck_port }}
+            failureThreshold: 1
```
madmatah: I find this value very low.
And since we have `replicas=1`, this leads to service interruption in case of false positive on the readiness check.

WDYT about setting `failureThreshold: 3` here ?

lunika (approved):

apps/etherpad/templates/services/app/dc.yml.j2
jmaupetit: I might be wrong, but this is a readiness probe, so there won't be any service interruption IMO since: i. in a rolling update scenario, the switch won't happen until the pod is ready, and, ii. in a blue-green scenario the deployment won't pursue if the pod is not ready (and the switch didn't happen yet).

FWIU, with a `failureThreshold` of 3, it means that we will need to wait for 3 failures to consider that the pod is not ready. Since we already know it the first time it occurs, why waiting for 2 other `periodsSeconds`?

madmatah: Oh, I didn't take in account the impact of the readiness probe during the deployment process while writing my comment.

From what I understand, the readiness probe has another impact at the service level  : if a pod is not ready, at anytime during its whole lifecycle, no traffic is sent to it by the service. That's why I said it generates an interruption.

Source : [K8s documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) : 

> A pod with containers reporting that they are not ready does not receive traffic through Kubernetes Services.

and 

> Note: Readiness probes runs on the container during its whole lifecycle.

jmaupetit: I see. Thanks for the tip :pray:

madmatah (approved):

PR #498, opened 2020-04-23
✨(core) add support for generic ConfigMap templates

jmaupetit:
## Purpose

Automagic ConfigMap generation is a useful feature to ease their creation from files that will be used mounted as volumes, but we need more flexibility by also allowing the creation of generic ConfigMap objects.

## Proposal

- [x] add support for ConfigMap creation from `cm*` service templates

### DIFF
--- a/tasks/get_objects_for_app.yml
+++ b/tasks/get_objects_for_app.yml
@@ -15,6 +15,7 @@
 - name: Set OpenShift objects to manage
   set_fact:
     builds: "{{ templates | map('regex_search', '.*/bc.*\\.yml\\.j2$') | select('string') | list }}"
+    configmaps: "{{ templates | map('regex_search', '.*/cm.*\\.yml\\.j2$') | select('string') | list }}"
     deployments: "{{ templates | map('regex_search', '.*/dc.*\\.yml\\.j2$') | select('string') | list }}"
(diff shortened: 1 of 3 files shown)


tasks/get_objects_for_app.yml
```diff
@@ -15,6 +15,7 @@
 - name: Set OpenShift objects to manage
   set_fact:
     builds: "{{ templates | map('regex_search', '.*/bc.*\\.yml\\.j2$') | select('string') | list }}"
+    configmaps: "{{ templates | map('regex_search', '.*/cm.*\\.yml\\.j2$') | select('string') | list }}"
```
lunika, line 18: This is not specific to this regex but I don't understand why the regex is not `.*/cm_.*\\.yml\\.j2$` ? Is `cmb.yml.j2` a valid filename ?

jmaupetit: Yes, `cmb.yml.j2` is a valid name. I don't enforce the usage of a underscore to allow a simple `cm.yml.j2` as we do for a deployment configuration, _e.g._ `dc.yml.j2`.

lunika (approved):

madmatah (approved):

PR #501, opened 2020-04-28
♻️(marsha) split app services to isolate xapi requests

lunika:
## Purpose

We want to split API and XAPI requests. With this split the API will
still be up while the XAPI part of the application have issues.

## Proposal

- [x] create dc, svc and nginx config dedicated to XAPI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Split API and XAPI requests in marsha
+
 ## [5.7.0] - 2020-04-23
 
 ### Added
--- a/apps/marsha/templates/services/app/svc_xapi.yml.j2
+++ b/apps/marsha/templates/services/app/svc_xapi.yml.j2
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: marsha
+    service: xapi
+    version: "{{ marsha_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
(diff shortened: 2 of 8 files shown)


apps/marsha/templates/services/app/svc_xapi.yml.j2
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: marsha
+    service: marsha
```
madmatah: Why not : 
```suggestion
    service: xapi
```

apps/marsha/templates/services/app/svc_xapi.yml.j2
```diff
@@ -0,0 +1,20 @@
...
+    version: "{{ marsha_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: marsha-xapi-{{ deployment_stamp }}  # name of the service should be host name in nginx
+  namespace: "{{ project_name }}"
+spec:
+  ports:
+  - name: 8000-tcp
+    port: 8000
+    protocol: TCP
+    targetPort: 8000
```
madmatah: Use `{{ marsha_django_port }}` instead of the hardcoded 8000 ?
It was already like this before your PR in the service definition, but I think it's an opportunity to improve this.

madmatah (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- split API and XAPI requests in marsha
```
```suggestion
- Split API and XAPI requests in marsha
```

jmaupetit (approved):
LGTM

PR #504, opened 2020-05-06
Refactor edxapp to remove collectstatic

sampaccoud:
## Purpose

We decided to pre-build edxapp images outside of Arnold. The pre-built images should be ready to run ie with applied custom theme if any and including static files as symlinks to avoid having to collectstatic files in a volume (cf https://github.com/openfun/openedx-docker/pull/223). 

As a result, we can completely remove everything related to building custom images and collecting static files from Arnold.

## Proposal

- [x] Remove static volumes
- [x] Remove collectstatic jobs
- [x] Remove git ssh key secret
- [x] Remove all build configurations and image streams
- [x] Remove macro to calculate image name
- [x] point to the base edxapp image


apps/edxapp/templates/services/nginx/dc.yml.j2
```diff
@@ -34,8 +32,12 @@ spec:
                   values:
                   - "edxapp-nginx-{{ deployment_stamp }}"
               topologyKey: kubernetes.io/hostname
+{% if edxapp_image_pull_secret_name is defined and edxapp_image_pull_secret_name is not none %}
```
lunika, line 35: Is it intended to not define it in `vars/all/main.yml` ?

sampaccoud: yes because there is this `if` clause so it does not need to be defined and the default is public repo?

lunika: Ok I understand

apps/edxapp/templates/services/cms/_dc_base.yml.j2
```diff
@@ -113,8 +113,6 @@ spec:
           name: edxapp-config
```
jmaupetit, line 113: Concerning your commit title, there is an emoji for reverting things in gitmoji: :rewind:

jmaupetit (approved):

apps/edxapp/vars/all/main.yml
```diff
@@ -7,7 +7,7 @@ edxapp_preview_host: "preview.{{ project_name}}.{{ domain_name }}"
 
 # -- edxapp (cms/lms)
 edxapp_image_name: "fundocker/edxapp"
-edxapp_image_tag: "hawthorn.1-oee-3.1.2"
+edxapp_image_tag: "hawthorn.1-oee-3.2.0"
```
sampaccoud: Actually i need to release a new edxapp image and pinning it here before merging this PR

apps/edxapp/vars/all/main.yml
```diff
@@ -112,7 +71,7 @@ edxapp_nginx_cms_admin_ip_withelist: []
 edxapp_nginx_cms_bypass_htaccess_ip_whitelist: []
 edxapp_nginx_lms_admin_ip_withelist: []
 edxapp_nginx_lms_bypass_htaccess_ip_whitelist: []
-edxapp_nginx_cms_static_cache_expires: "1M"
+edxapp_nginx_cms_static_cache_expires: "1d"
```
sampaccoud, line 74: Actually I should include this in a separate commit and explain that Open edX does not version static files that are called via `require` (they may change when we upgrade the base open edx image of one of our site). So we should not set cache expiries too long.

sampaccoud: done

PR #505, opened 2020-05-06
✨(marsha) create custom settings module

lunika:
## Purpose

In Marsha we want to redefine settings module if needed. We just created
a settings module importing every settings and the good one will be used
thanks to the DJANGO_CONFIGURATION environment variable

## Proposal

- [x] create custom settings module

### DIFF
--- a/apps/marsha/templates/services/app/_dc_base.yml.j2
+++ b/apps/marsha/templates/services/app/_dc_base.yml.j2
@@ -58,7 +58,7 @@ spec:
             periodSeconds: 5
           env:
             - name: DJANGO_SETTINGS_MODULE
-              value: marsha.settings
+              value: marsha.configs.settings
             - name: DJANGO_CONFIGURATION
               value: "{{ marsha_django_configuration }}"
             - name: POSTGRES_DB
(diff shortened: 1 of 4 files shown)


apps/marsha/templates/services/app/_dc_base.yml.j2
```diff
@@ -58,7 +58,7 @@ spec:
             periodSeconds: 5
           env:
             - name: DJANGO_SETTINGS_MODULE
-              value: marsha.settings
+              value: marsha.custom_settings.settings
```
jmaupetit:
```suggestion
              value: marsha.configs.settings
```

apps/marsha/templates/services/app/_dc_base.yml.j2
```diff
@@ -77,6 +77,8 @@ spec:
             - configMapRef:
                 name: "marsha-app-dotenv-{{ deployment_stamp }}"
           volumeMounts:
+            - name: marsha-configmap
+              mountPath: /app/src/backend/marsha/custom_settings
```
jmaupetit:
```suggestion
              mountPath: /app/src/backend/marsha/configs
```

jmaupetit (approved):
Great! I've added a suggestion for the custom settings module path (similar to edxapp).

PR #506, opened 2020-05-06
✨(redis-sentinel) allow to disable persistence

lunika:
## Purpose

Persistence in redis-sentinel is now optional and the variable
redis_sentinel_redis_persistence_enabled is here to enable it or not. By
default this variable is set to false.

## Proposal

- [x] allow to disable persistence

### DIFF
--- a/apps/redis-sentinel/templates/services/app/svc.yml.j2
+++ b/apps/redis-sentinel/templates/services/app/svc.yml.j2
@@ -12,7 +12,7 @@ metadata:
     version: RedisFailover
   # Nota bene: the name of the service should be the host name in edxapp configuration,
   # prefixed by "rfs-", e.g. "rfs-redis-sentinel"
-  name: redis-sentinel
+  name: {{ redis_sentinel_service_name }}
   namespace: "{{ project_name }}"
 spec:
   sentinel:
@@ -59,6 +59,7 @@ spec:
       - "{{ config }}"
 {% endfor %}
 {% endif %}
+{% if redis_sentinel_redis_persistence_enabled %}
     storage:
       keepAfterDeletion: true
       persistentVolumeClaim:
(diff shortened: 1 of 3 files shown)


apps/redis-sentinel/templates/services/app/svc.yml.j2
```diff
@@ -59,6 +59,7 @@ spec:
       - "{{ config }}"
 {% endfor %}
 {% endif %}
+{% if redis_sentinel_redis_persistence_enabled %}
```
jmaupetit, line 62: WDYT?
```suggestion
{% if redis_sentinel_redis_persistence_is_enabled %}
```

apps/redis-sentinel/vars/all/main.yml
```diff
@@ -6,6 +6,10 @@ redis_sentinel_user_id: null
 redis_sentinel_group_id: null
 redis_sentinel_fsgroup_id: null
 
+## Service fonciguration
```
jmaupetit:
```suggestion
## Service configuration
```

apps/redis-sentinel/templates/services/app/svc.yml.j2
lunika: I'm not a big fan of the `is` suffix, `enabled` is enough for me but why not.

jmaupetit: ok then

jmaupetit (approved):

madmatah (approved):
~~Is it possible to mention somewhere that redis will still write on a temporary disk if the persistence is disabled ?~~

~~By `disable persistence`, one could think that it is in-memory only.~~


Sorry I removed my comment, not relevant.
And this is consistent with redis-operator semantics.

PR #511, opened 2020-05-20
Add Kibana app

jmaupetit:
## Purpose

Kibana + ES is often used to perform data analytics. We need to be able to add this service to our projects.

## Proposal

- [x] add `StatefulSet` objects support
- [x] add Elasticsearch 7.0+ support
- [x] add Kibana application


apps/elasticsearch/templates/services/app/dc.yml.j2
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-kind: DeploymentConfig
+kind: StatefulSet
```
jmaupetit, line 2: Maybe we should create a new type of objects that Arnold should support, but IMO a `StatefulSet` is a kind of deployment...

jmaupetit: After discussing with @madmatah we decided to create an explicit object type with its own life cycle, or else, these objects would never get cleaned.

apps/kibana/templates/services/app/configs/kibana.yml.j2
```diff
@@ -0,0 +1,11 @@
+# Arnold-controlled configuration
+
+# The server name should be unique for all replica
+server.name: ${HOSTNAME}
+
+# Accept request from all hosts
```
jmaupetit:
```suggestion
# Accept requests from all hosts
```

apps/kibana/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,67 @@
...
+        - name: kibana
+          image: "{{ kibana_image_name }}:{{ kibana_image_tag }}"
+          imagePullPolicy: Always
+          livenessProbe:
+            httpGet:
+              path: /api/status
+              port: {{ kibana_app_port }}
+              httpHeaders:
+                - name: Host
+                  value: "{{ kibana_host }}"
```
jmaupetit: This header is not required.

apps/kibana/templates/services/app/dc.yml.j2
```diff
@@ -0,0 +1,67 @@
...
+            initialDelaySeconds: 60
+            periodSeconds: 30
+          readinessProbe:
+            httpGet:
+              path: /api/status
+              port: {{ kibana_app_port }}
+              httpHeaders:
+                - name: Host
+                  value: "{{ kibana_host }}"
```
jmaupetit: Ditto.

apps/kibana/templates/services/nginx/configs/kibana.conf.j2
```diff
@@ -0,0 +1,49 @@
...
+{% set bypass_htaccess = kibana_nginx_bypass_htaccess_ip_whitelist | length > 0 %}
+upstream kibana-backend {
+  server kibana-app-{{ deployment_stamp }}:{{ kibana_app_port }} fail_timeout=0;
+}
+
+server {
+  listen {{ kibana_nginx_port }};
+  server_name localhost;
+
+  client_max_body_size 100M;
```
madmatah: Can you please add comments about this `100M` choice ? This is huge !

And I don't think upstream load balancer(s) will allow this without configuration change.

apps/kibana/vars/vault/main.yml.j2
```diff
@@ -0,0 +1,7 @@
+# customer: {{ customer }}
+# env_type: {{ env_type }}
+
+# kibana
+XPACK_SECURITY_ENCRYPTIONKEY: {{ lookup('password', '/dev/null length=50') }}
+XPACK_REPORTING_ENCRYPTIONKEY: {{ lookup('password', '/dev/null length=50') }}
+XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY: {{ lookup('password', '/dev/null length=50') }}
```
madmatah, line 7: From what I understand, you provides random generated passwords as default values for these secrets.
The goal is that each customer define its own secrets in a vault file.

But if you forget to define these for a customer, these default values will be applied silently, with no warning.
And a side effect that I see is that the secrets will change each time the playbook is applied.

According to the [ansible documentation ](https://docs.ansible.com/ansible/latest/plugins/lookup/password.html): 
> A special case is using /dev/null as a path. The password lookup will generate **a new random password each time**

jmaupetit: This file is a vault template used once while generating the application's vault for your project using the `create_vaults.yml` playbook. When the vault has been created with random generated passwords, it can be edited using the `ansible-vault` tool but it won't be changed when generating secrets.

madmatah: Ok, I missed the fact that it is used only in the`create_vaults.yml` playbook !
Thanks for the clarifications :+1:

apps/kibana/templates/services/nginx/configs/kibana.conf.j2
jmaupetit: This was a copy/pasta fail. I'll remove it.

jmaupetit:
@madmatah ready for another (final) review! :pray:

madmatah (approved):

PR #516, opened 2020-05-28
🐛(core) fix openshift-acme role permissions

jmaupetit:
## Purpose

Recent `openshift-acme` releases require more permissions to boot.

## Proposal

- [x] fix application role template


sampaccoud (approved):
LGTM! I would just question the change to single quotes :man_shrugging:

jmaupetit:
> I would just question the change to single quotes

No one is allowed to question Prettier. No one. :trollface:

PR #527, opened 2020-07-23
🐛(core) exclude acme routes when looking for app routes

lunika:
## Purpose

When using acmev2 challenge routes contains the label included by the
targeted route. We have to exclude acme route to have only the next,
current or previous route returned by k8s.

## Proposal

- [x] exclude acme routes when looking for app routes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
+- Exclude `openshift-acme` routes when looking for app routes
+
 ## [5.13.0] - 2020-07-22
 
 ### Added
--- a/tasks/deploy_get_stamp_from_route.yml
+++ b/tasks/deploy_get_stamp_from_route.yml
@@ -14,6 +14,7 @@
     label_selectors:
       - app={{ app.name }}
       - route_prefix={{ prefix }}
+      - "!acme.openshift.io/temporary"
   register: app_route
   when: routes | length > 0 and app.settings.is_blue_green_compatible | default(True)
 


tasks/deploy_get_stamp_from_route.yml
```diff
@@ -11,9 +11,7 @@
     api_version: "v1"
     namespace: "{{ project_name }}"
     kind: "Route"
-    label_selectors:
-      - app={{ app.name }}
-      - route_prefix={{ prefix }}
+    label_selectors: ["app={{ app.name }},route_prefix={{ prefix }},!acme.openshift.io/temporary"]
```
sampaccoud: curious why not use the normal list syntax?

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
+- exclude acme routes when looking for app routes
```
sampaccoud:
```suggestion
- exclude acme routes when looking for app routes in the blue/green switch
```

tasks/deploy_get_stamp_from_route.yml
lunika: I have a syntax error when adding a new entry with `-!acme.openshift.io/temporary`. We alreay use this syntax here https://github.com/openfun/arnold/blob/master/tasks/manage_app.yml#L101

sampaccoud: Interesting. Maybe surrounding it with quotes would also work.

CHANGELOG.md
lunika: This part of the code is not used only by the switch task. We also use it in the clean, rollback, deploy.

```suggestion
- Exclude `openshift-acme` routes when looking for app routes
```

jmaupetit (approved):

tasks/deploy_get_stamp_from_route.yml
lunika: It works using quotes.

lunika: I changed with the previous syntax and I use quotes now.

PR #530, opened 2020-07-31
✨(apps) choose to enable acme based on route prefix

lunika:
## Purpose

With acme v2, each time it tries to resolve a challenge a new route is
    created and an associated pod is also created. In some cases, routes
    next and previous exist but are not used, DNS is not created for example,
    the challenge is never resolved and routes and pods are still living in
    the project consuming resources for nothing. We introduced a new
    variable `acme_enabled_route_prefix` allowing to choose on which route
    acme will generate a certificate based on their prefix.

I also made a command to patch existing routes : 

```
oc patch route $(oc get routes --output=json $(oc get routes | grep "exposer" | awk '{print $1}') | jq -r '.items[].metadata.annotations."acme.openshift.io/exposer-key"' | cut -d "/" -f 2) --type=json -p='[{"op": "remove", "path": "/metadata/annotations/kubernetes.io~1tls-acme"}, {"op": "remove", "path": "/metadata/annotations/acme.openshift.io~1status"}]'
```

patch is made on route having an associated exposer route.


jmaupetit (commented):
I would prefer a more generic solution such as defining a variable that disables acme for named routes, such as:

```yaml
# core_apps/acme/vars/all/main.yml
# [...]
acme_disabled_routes:
  - ashley-nginx-next
  - forum-previous
```

jmaupetit:
> This PR if for now a PoC, do you see some unwanted behaviour if the kubernetes.io/tls annotation is removed ?

If you mean `kubernetes.io/tls-acme`, no.

apps/edxapp/templates/services/nginx/route_cms.yml.j2
```diff
@@ -12,7 +12,9 @@ metadata:
     route_prefix: "{{ prefix }}"
     route_target_service: "cms"
   annotations:
+{% if prefix == 'current' or edxapp_nginx_disable_acme_on_next_and_previous is false %}
```
jmaupetit: Implementation would look like:

```suggestion
{% if 'edxapp-nginx-cms-' + prefix not in acme_disabled_routes %}
```

jmaupetit:
> I would prefer a more generic solution such as defining a variable that disables acme for named routes, such as:
> 
> ```yaml
> # core_apps/acme/vars/all/main.yml
> # [...]
> acme_disabled_routes:
>   - ashley-nginx-next
>   - forum-previous
> ```

Or alternative and less error-prone approach may be to list app/service/prefix routes explicitly:

```yaml
# core_apps/acme/vars/all/main.yml
# [...]
acme_disabled_routes:
  - app: ashley
    service: nginx
    prefix: next
  - app: forum
    service: app
    prefix: previous
```

group_vars/all/main.yml
```diff
@@ -78,6 +78,11 @@ acme_env: staging
 # need to create it for each project. Hence set "has_acme_cluster_role" to
 # "true"
 has_acme_cluster_role: false
+# Which route prefix should have a certificate generated by acme
+acme_enabled_route_prefix:
+  - current
+  - previous
```
jmaupetit, line 84:
```suggestion
  - previous
  - current
```

jmaupetit (approved):

PR #534, opened 2020-08-18
🐛(core) create only required acme issuer CM

jmaupetit:
## Purpose

As mentioned in `openshift-acme`'s documentation:

> Let's encrypt provides two environments: live and staging. The environment is chosen based on the issuer ConfigMap that is created.

Hence, when creating both live and staging CM issuers, on particular cases, `openshift-acme` can issue staging certificates for production environments.

## Proposal

- [x] only create required issuer CM given the environment

Hat tip to @lunika for this suggestion. :tophat: :muscle:


core_apps/acme/templates/services/app/cm.yml.j2
```diff
@@ -0,0 +1,15 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: letsencrypt
+  annotations:
+    "acme.openshift.io/priority": {% if acme_env == "live" %}"100"{% else %}"50"{% endif %}
```
madmatah: I don't think that this test is needed, since the priority annotation is only used to [sort a list](https://github.com/tnozicka/openshift-acme/blob/8b54945f9457b1b960fea133bce359679eee5d62/pkg/controllerutils/controllerutils.go#L51-L89) if multiple CertIssuer are defined.

```suggestion
    "acme.openshift.io/priority": 100
```

jmaupetit: True.

madmatah (approved):

PR #535, opened 2020-08-19
✨(richie) allow placing static files behind a CDN

sampaccoud:
## Purpose

On richie sites, we can configure Django to serve static files behind a CDN like CloudFront on a domain other then the origin.
Some types of files like fonts required CORS access control allow origin headers to be set.

## Proposal

Add minimal CORS headers to requests on static files so they can be served from another domain name as long as the site origin url is our main domain.

### DIFF
--- a/apps/richie/templates/services/nginx/configs/richie.conf.j2
+++ b/apps/richie/templates/services/nginx/configs/richie.conf.j2
@@ -84,5 +84,11 @@ server {
     add_header Cache-Control public;
     root /data/static;
     try_files /$file =404;
+
+    add_header Access-Control-Allow-Origin https://{{ richie_host }};
+    add_header Access-Control-Allow-Credentials false;
+    add_header Access-Control-Allow-Methods GET;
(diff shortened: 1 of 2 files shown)


apps/richie/templates/services/nginx/configs/richie.conf.j2
```diff
@@ -84,5 +84,11 @@ server {
     add_header Cache-Control public;
     root /data/static;
     try_files /$file =404;
+
+    add_header Access-Control-Allow-Origin https://{{ richie_host }};
```
jmaupetit, line 88: Should we enforce the `https` protocol? This breaks HTTP compatibility.

sampaccoud: I thought it was already the case in our deployments. Where can I get this info from variables?

lunika: As far as I know we have no variable containing which schema to use. But all our routes are configured to use `https` I think.

jmaupetit: From a route perspective, you are right: https://github.com/openfun/arnold/blob/master/apps/richie/templates/services/nginx/route.yml.j2#L22 I was focused on the nginx configuration.

jmaupetit (approved):

PR #536, opened 2020-08-21
♻️(richie) inject configmap environment in all jobs to allow defining required settings

sampaccoud:
## Purpose

We want to add an environment variable to our configmap for a setting that is required. Our jobs are breaking because the configmap env is not injected.

## Proposal

Modify the template definition for all jobs to inject the configmap env.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


lunika (approved):

sampaccoud:
@madmatah I have second thoughts on this as we can just add the setting to secrets (which is already overriden in our richie projects deployments...). Do you think it is not appropriate?

PR #542, opened 2020-09-03
🐛(playbook) fix the `delete_app` task to avoid timeout

madmatah:
## Purpose

When executing a `clean` on a project having `Cronjob` objects, it fails with a timeout on the step `Wait for objects of kind Pod to be deleted`.

The reason is that we are trying to delete pods before cronjobs.
As soon as we delete a pod related to a cronjob (or a job), it is automatically re-created.
And we can wait forever for the pods to be deleted.

## Proposal

- [x] Delete pods after all kind of objects that can spawn new pods


tasks/delete_app.yml
```diff
@@ -11,9 +11,9 @@
     # https://github.com/ansible/ansible/issues/42302
     - { kind:  "ReplicationController", api_version: "v1" }
     - { kind:  "StatefulSet", api_version: "v1" }
-    - { kind:  "Pod", api_version: "v1" }
     - { kind:  "Job", api_version: "v1" }
     - { kind:  "CronJob", api_version: "v1beta1" }
+    - { kind:  "Pod", api_version: "v1" }
```
sampaccoud, line 17: this PR is beautiful :star_struck:  add a comment to secure it?

sampaccoud (approved):

tasks/delete_app.yml
madmatah: Thanks! :slightly_smiling_face: 

I'll add a comment.
If you don't mind, i'll also add a commit to bump arnold version in the same PR to avoid running the CI twice.

sampaccoud: oh yes!

madmatah: @sampaccoud Done !

lunika (approved):

sampaccoud (approved):

jmaupetit (approved):

PR #548, opened 2020-09-22
Circle-ci optimization

slandrault:
## Purpose

In order to improve CI performances, this PR will introduce the concept of "selective ci", similar to what was done for https://github.com/openfun/openedx-docker. The main improvement will be to reduce the number of jobs run to only apps that were modified or that are required by another job.

## Proposal

To do so, some CI refactoring has been done:

- All CI jobs are now described in the `.circleci/templates` folder. Each subfolder represent a job (folder name = job name) and inside each folder will be a `job.yml` file to describe the job itself and a `workflow.yml` file to describe the job in the arnold workflow (requirements, filters...)
- The `config.yml` will be autogenerated according to what has changed in *global* files (that may impact every apps) and according to apps themselves (an app with no change won't be tested). This config file will be created with a call to `bin/update_config` script (that will then call `.circleci/scripts/update_config.py` script). The developer pushing a new commit must run it before commiting.
- The `config.yml` will be verified by the ci itself to make sure that it is up to date and building proper project.

### DIFF
--- a/.circleci/scripts/update_config.py
+++ b/.circleci/scripts/update_config.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import subprocess
+import textwrap
+import yaml
+
+
+CI_PATH = '.circleci'
+CI_ROOT = CI_PATH + '/templates'
+CI_CONFIG = CI_PATH + '/config.yml'
+APPS_ROOT = 'apps'
+ROOT_FILES = [  # If any of those file/folder content changed, build all known jobs
+    'env.d/',
+    'filter_plugins/',
+    'group_vars/',
+    'lookup_plugins/',
+    'tasks/',
+    'templates/',
+    'tests/',
+    # + any file that is root
+]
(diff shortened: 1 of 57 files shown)


.circleci/scripts/update_config.py
```diff
@@ -0,0 +1,139 @@
...
+
+import os
+import re
+import subprocess
+import yaml
+
+
+def indent(text, indent=4):
+    """indent a text with n spaces"""
+    return '\n'.join('{}{}'.format(' ' * indent, line) for line in text.split('\n'))
```
jmaupetit: Do you know the `textwrap` module [1]? I discovered this recently.

[1] https://docs.python.org/3/library/textwrap.html#textwrap.indent

slandrault: I did not! Thx for the tip!

jmaupetit:
Closing this as our first intent was to add support for application trays (see #309) and move trays testing to related projects.

PR #567, opened 2020-11-06
✨(marsha) add a cronjob to stop unused live streaming

lunika:
## Purpose

Marsha contains a management command checking all live streaming and stop them if they are unused. 

To use this management command we created a cronjob scheduled every 20 minutes running this management command

## Proposal

- [x] add a cronjob to stop unused live streaming

### DIFF
--- a/apps/marsha/templates/services/app/cronjob_check_live_state.yml.j2
+++ b/apps/marsha/templates/services/app/cronjob_check_live_state.yml.j2
@@ -1,0 +1,64 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  labels:
+    app: marsha
+    service: app
+    version: "{{ marsha_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: "check-live-state-{{ deployment_stamp }}"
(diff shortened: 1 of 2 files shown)


apps/marsha/templates/services/app/cronjob_check_live_state.yml.j2
```diff
@@ -0,0 +1,87 @@
...
+                claimName: marsha-pvc-media
+            - name: marsha-v-static
+              persistentVolumeClaim:
+                claimName: marsha-pvc-static
+{% endif %}
+{% if marsha_should_sign_requests %}
+            - name: marsha-cloudfront-private-key-secret
+              secret:
+                secretName: "{{ marsha_cloudfront_private_key_secret_name }}"
+{% endif %}
```
sampaccoud: I wouldn't mount static, media and cloudfront private key...

sampaccoud (approved):

apps/marsha/templates/services/app/cronjob_check_live_state.yml.j2
lunika: I'm not sure it will work without cloudfront private key

PR #574, opened 2020-11-18
[edxapp] Set "Referer" header with "Origin" header when empty and release fix

sampaccoud:
## Fixed

- Set Referer header with Origin header when empty in edxapp's lms nginx


apps/edxapp/templates/services/nginx/configs/lms.conf.j2
```diff
@@ -4,6 +4,12 @@ upstream lms-backend {
   server edxapp-lms-{{ deployment_stamp }}:{{ edxapp_django_port }} fail_timeout=0;
 }
 
+# Set refere header with origin header when empty
```
lunika:
```suggestion
# Set refer header with origin header when empty
```

lunika (approved):

apps/edxapp/templates/services/nginx/configs/lms.conf.j2
lunika: Oh my bad it's `referer` not `refer` sorry.

PR #576, opened 2020-11-25
Add arnold CLI

jmaupetit:
## Purpose

If we want to easily package, distribute and deploy Arnold along with its
applications (aka trays), we need a command line tool dedicated to the task.

## Proposal

- [x] remove legacy scripts
- [x] improve Docker image for production
- [x] add `arnold` CLI
- [x] use the CLI in the CI
- [x] ~update documentation (may be updated in a new PR)~ postponed to #582


bin/arnold
```diff
@@ -0,0 +1,818 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+# Docker
+declare -r ARNOLD_IMAGE_NAME="${ARNOLD_IMAGE_NAME:-fundocker/arnold}"
+declare -r ARNOLD_IMAGE_TAG="${ARNOLD_IMAGE_TAG:-5.21.0}"
```
jmaupetit: Should be up-to-date.

bin/activate
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+export ARNOLD_IMAGE_NAME="arnold"
+ARNOLD_IMAGE_TAG="$(tr -d '\n' < VERSION)"
+export ARNOLD_IMAGE_TAG
+export ARNOLD_IMAGE="${ARNOLD_IMAGE_NAME}:${ARNOLD_IMAGE_TAG}"
+export ANSIBLE_VAULT_PASSWORD="arnold"
+export K8S_AUTH_VERIFY_SSL="no"
+
+echo -e "Development environment activated ✅\\n"
```
madmatah, line 30: Maybe this script want to check that it has been "sourced" ?
Because when executed directly with `bin/activate`, it has no effect.

bin/activate
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+export ARNOLD_IMAGE_NAME="arnold"
+ARNOLD_IMAGE_TAG="$(tr -d '\n' < VERSION)"
+export ARNOLD_IMAGE_TAG
+export ARNOLD_IMAGE="${ARNOLD_IMAGE_NAME}:${ARNOLD_IMAGE_TAG}"
+export ANSIBLE_VAULT_PASSWORD="arnold"
```
madmatah: The value of the default password (`"arnold"`) is duplicated 3 times in the project : 
 - bin/activate
 - Makefile
 - CircleCI config

Should it be defined once in a file?

Makefile
```diff
@@ -0,0 +1,116 @@
+# -- ARNOLD
+ARNOLD_IMAGE_NAME      = arnold
+ARNOLD_IMAGE_TAG       = $(shell tr -d '\n' < VERSION)
+ARNOLD_IMAGE_TAG_DEV   = $(ARNOLD_IMAGE_TAG)-dev
+ARNOLD_IMAGE           = $(ARNOLD_IMAGE_NAME):$(ARNOLD_IMAGE_TAG)
+ARNOLD_IMAGE_DEV       = $(ARNOLD_IMAGE_NAME):$(ARNOLD_IMAGE_TAG_DEV)
+ANSIBLE_VAULT_PASSWORD = arnold
```
madmatah: Should it use environment variables instead of these values, when defined ?

```suggestion
ARNOLD_IMAGE_NAME      ?= arnold
ARNOLD_IMAGE_TAG       ?= $(shell tr -d '\n' < VERSION)
ARNOLD_IMAGE_TAG_DEV   ?= $(ARNOLD_IMAGE_TAG)-dev
ARNOLD_IMAGE           ?= $(ARNOLD_IMAGE_NAME):$(ARNOLD_IMAGE_TAG)
ARNOLD_IMAGE_DEV       ?= $(ARNOLD_IMAGE_NAME):$(ARNOLD_IMAGE_TAG_DEV)
ANSIBLE_VAULT_PASSWORD ?= arnold
```

Makefile
```diff
@@ -0,0 +1,116 @@
...
+ARNOLD_RUN_DEV = $(ARNOLD) -d -- run
+
+# -- Docker
+# Get the current user ID to use for docker run and docker exec commands
+DOCKER_UID  = $(shell id -u)
+DOCKER_GID  = $(shell id -g)
+DOCKER_USER = $(DOCKER_UID):$(DOCKER_GID)
+
+# -- k8s
+K8S_DOMAIN="$(shell hostname -I | awk '{print $$1}')"
```
madmatah:
```suggestion
K8S_DOMAIN?="$(shell hostname -I | awk '{print $$1}')"
```

bin/arnold
```diff
@@ -0,0 +1,832 @@
...
+
+# Hosts
+
+# K8S_DOMAIN is only used in development and the CI to set the project's
+# 'domain_name' to "${K8S_DOMAIN}.nip.io"
+declare K8S_DOMAIN="${K8S_DOMAIN:-$(hostname -I | awk '{print $1}')}"
+
+# Flags
+declare -i DEV_MODE=0
+declare -i USE_GPG=0
```
jmaupetit: We must be able to bypass this default thanks to the environment:

```suggestion
declare -i USE_GPG=${ARNOLD_USE_GPG:-0}
```

bin/arnold
```diff
@@ -0,0 +1,832 @@
...
+    fi
+
+    # Get K8S credentials
+    export K8S_DOMAIN
+    export K8S_SA_TOKEN
+
+    K8S_AUTH_HOST=$(_get_k8s_auth_host)
+    export K8S_AUTH_HOST
+
+    # Check that the current user is already logged in k8s
```
madmatah:
```suggestion
    # Check that the current user is already logged in k8s. Note that the `whoami`
    # command is only compatible with openshift and should be adapted for k8s.
```

bin/arnold
```diff
@@ -0,0 +1,832 @@
...
+        log info "Development mode activated for tray: ${ARNOLD_TRAY_NAME}"
+      else
+        extras_volumes="-v ${PWD}:/app"
+        log info "Full development mode activated"
+      fi
+    fi
+
+    # shellcheck disable=SC2086
+    docker run --rm -ti \
+        -u "$(id -u)" \
```
madmatah: What do you think about adding a variable like this : 
```
DOCKER_USER="${DOCKER_USER:-$(id -u)}"
```

and use it as a value for the `-u` argument  ?

bin/activate
jmaupetit: I've added a warning message and an error code while trying to execute this script.

bin/arnold
```diff
@@ -0,0 +1,829 @@
...
+
+# ---- Utils library ----
+
+# Compute K8S_AUTH_HOST from its host
+function _get_k8s_auth_host() {
+    ${KUBECTL} version | grep Server | awk '{print $2}'
+}
+
+function login() {
+    k8s_auth_host=$(_get_k8s_auth_host)
```
lunika, line 294: You can't have the host if you are not logged in.

This is the output I have when I'm not connected :

```
$ oc version
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO
```

bin/arnold
```diff
@@ -0,0 +1,829 @@
...
+    # Allow overriding config per customer
+    customer_vars="$(dirname "${BASH_SOURCE[0]}")/../${CUSTOMER_DIR}/${ARNOLD_CUSTOMER}/vars.inc.sh"
+    if [ -e "${customer_vars}" ]; then
+        # shellcheck source=/dev/null
+        source "${customer_vars}"
+    fi
+
+    # Get K8S credentials
+    export K8S_DOMAIN
+    export K8S_SA_TOKEN
```
lunika: I don't understand these 2 export. Where are they set ? What is exported ? Do we really need to export them ?

bin/arnold
```diff
@@ -0,0 +1,829 @@
...
+
+
+# ---- Utils library ----
+
+# Compute K8S_AUTH_HOST from its host
+function _get_k8s_auth_host() {
+    ${KUBECTL} version | grep Server | awk '{print $2}'
+}
+
+function login() {
```
lunika: This function is not listed in the usage function.

bin/arnold
```diff
@@ -0,0 +1,829 @@
...
+        if [[ -z "${K8S_SA_TOKEN}" ]] ; then
+            echo "Error: you need to login to a k8s server first."
+            exit 1
+        else
+            login
+        fi
+    fi
+
+    # Ansible's k8s raw module requires the following two environment
+    # variables to be defined
```
lunika: There is only one defined

.gitignore
```diff
@@ -16,6 +16,7 @@ env.d/*
 # Tests temporary files
 .coverage
 .pytest_cache
+.pylint.d
```
lunika, line 19: You should complete your description explaining why you use `-d` and what is `-d` option.

.circleci/config.yml
```diff
@@ -83,23 +80,17 @@ version: 2
...
-
-    working_directory: ~/fun
+    <<: *defaults
 
     steps:
       - checkout
-      - *ci_env
       - *docker_login
       - run:
           name: Build arnold production image
```
lunika: The name is not accurate anymore with what you are doing

bin/activate
jmaupetit: I'm not sure this is good idea as people might use it without knowing they are using a default vault password. In fact those occurrences in 3 different files serve 3 different (undocumented) purposes:

- the `bin/activate` script is used to activate a development environment to use with the `bin/arnold` script,
- the `Makefile` is only used for development purpose with no other dependencies (see next comment), and,
- the `.circleci/config.yml` file is used for CI usage only.

That said, I'm open to suggestions, what were you thinking of by "defined once in a file"?

Makefile
jmaupetit: This is what I did in a first attempt, and, I changed my mind to enforce default development values. Do you have a particular scenario in mind where it could be useful to redefine those thanks to the environment?

Makefile
jmaupetit: Ok, maybe this one would be useful :sweat_smile:

bin/arnold
jmaupetit: Well spotted! I've changed to the following:

```bash
function _get_k8s_auth_host() {
    echo "https://${K8S_DOMAIN}:8443"
}
```

bin/arnold
jmaupetit: Those are not required. I've cleaned this. :pray:

bin/arnold
jmaupetit: It is not supposed to get called as requiring a service account token, I made it private.

bin/arnold
jmaupetit: Both `K8S_AUTH_API_KEY` and `K8S_AUTH_HOST` environment variables should be defined. I've slightly update my previous change.

.gitignore
jmaupetit: It should be mentioned in the documentation (README) or do you have something else in mind?

.circleci/config.yml
jmaupetit: Do you suggest making it plural? What about:

> Build arnold production & development images

jmaupetit:
@lunika & @madmatah I've addressed all your comments (thanks for the review) and pushed fixup commits to ease the second round. Ready?

bin/arnold
jmaupetit: In fact, this is breaking the CI as the DOCKER_USER already exists (as a secret)... :sweat_smile:

.gitignore
lunika: I mean commit description

.circleci/config.yml
lunika: This is more accurate

.gitignore
jmaupetit: I propose the following:

```
(project) use arnold -d run to execute dev commands

Instead of having x scripts to run development commands, use a unique
entrypoint: bin/arnold using the -d option to activate the `DEV_MODE`.

An active `DEV_MODE` implies the following behavior:

- `docker run` commands mount the local directory sources as a volume in `/app`, or,
- `docker run` commands mount the local application tree in `/app/apps/myapp` if a tray as been found in the repository.

Makefile rules have been added to run tests and linters.
```

WDYT?

lunika: ok for me

Makefile
madmatah: One thing that comes to my mind is that CircleCI is using `make` in its jobs. And also, it defines these some of these environment variables in `ci_env` task. At this time, the values of these environment variables are the same than the default values defined in the Makefile.

But if someone wants to change the variables defined in `ci_env` job in the future, I'm pretty sure that this will lead to unexpected behavior and confusion.

bin/activate
madmatah: I was thinking something like this:

------------

1) Add a `arnold_defaults` file containing : 
```
export ARNOLD_DEFAULT_VAULT_PASSWORD="arnold"
```

2) In the Makefile

```
include arnold_defaults

ANSIBLE_VAULT_PASSWORD=$(ARNOLD_DEFAULT_VAULT_PASSWORD)
```

3) In `bin/activate` : 

```
source arnold_defaults
export ANSIBLE_VAULT_PASSWORD="${ARNOLD_DEFAULT_VAULT_PASSWORD}"
```

4) In .circleci/config.yml : 

```
cat arnold_defaults >> $BASH_ENV
echo 'export ANSIBLE_VAULT_PASSWORD="${ARNOLD_DEFAULT_VAULT_PASSWORD}"' >> $BASH_ENV
```

------------

It avoids duplicating the value in the project. And about that :

> I'm not sure this is good idea as people might use it without knowing they are using a default vault password.

I don't think this suggestion makes the problem worse.
If it's a problem, you should print a message saying «Warning, you are using the default arnold password, your vault is not safe».

jmaupetit: Great suggestion! I'll do this. Thanks.

bin/arnold
```diff
@@ -37,6 +47,8 @@ declare -r CUSTOMER_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_CUSTOMER_DIR}"
 declare -r VAULT_PASSWORD_FILE_NAME="password.gpg"
 
 # Binaries
+declare -r CURL="docker run --rm curlimages/curl"
```
lunika: No please, don't use docker to run `curl`. Use directly `curl` or `wget`, `wget` is installed everywhere.

jmaupetit: > wget is installed everywhere.

That's a strong statement. I'm not sure to share this thought.

What about trying to use locally installed tools and fall back to the docker image?

Makefile
jmaupetit: Ok then.

bin/arnold
jmaupetit: Ok then. After an IRL discussion, `curl` is now a project dependency. Arnold will fail with an explicit error message if it's not installed.

jmaupetit:
@lunika & @madmatah I've addressed all your comments. Ready for a new review. :pray:

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+declare -r RELATIVE_COMMON_DIR="group_vars/common"
+declare -r RELATIVE_CUSTOMER_DIR="group_vars/customer"
+declare -r COMMON_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_COMMON_DIR}"
+declare -r CUSTOMER_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_CUSTOMER_DIR}"
+
+# Files
+declare -r VAULT_PASSWORD_FILE_NAME="password.gpg"
+
+# Dependencies
+declare -r CURL="curl"
```
lunika: I don't see the interest of this variable.

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+declare -r ARNOLD_PROJECT_PATH="${PWD}"
+declare -r ARNOLD_TRAY_FILE="arnold.yml"
+declare -r DOCKER_USER="${DOCKER_USER:-$(id -u)}"
+
+# Arnold release info
+declare LATEST_RELEASE
+declare LATEST_VERSION
+
+# Script infos
+declare SCRIPT_NAME
```
lunika, line 18: Why not declaring and assigning at the same time, like you do after in the script ?

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+SCRIPT_NAME=$(basename "$0")
+declare SCRIPT_PATH
```
lunika, line 20: devito

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+  ["notset"]=${LOG_LEVEL_NOTSET}
+)
+declare -A LOG_LEVEL_COLORS=(
+  ["critical"]=${COLOR_RED}
+  ["error"]=${COLOR_RED}
+  ["warning"]=${COLOR_ORANGE}
+  ["success"]=${COLOR_GREEN}
+  ["info"]=${COLOR_BLUE}
+  ["debug"]=${COLOR_DEFAULT}
+  ["notset"]=${COLOR_DEFAULT}
```
lunika, line 66: `notset` ? Kesako ?

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+  customer   : ${COLOR_BLUE}${ARNOLD_CUSTOMER}${COLOR_RESET}
+  environment: ${COLOR_BLUE}${ARNOLD_ENVIRONMENT}${COLOR_RESET}
+  gpg enabled: ${COLOR_BLUE}${USE_GPG}${COLOR_RESET}
+  dev mode   : ${COLOR_BLUE}${DEV_MODE}${COLOR_RESET}
+  log level  : ${COLOR_BLUE}$(_get_log_level_name)${COLOR_RESET}
+  CLI release: ${COLOR_BLUE}${ARNOLD_IMAGE_TAG}${COLOR_RESET}
+  image      : ${COLOR_BLUE}${ARNOLD_IMAGE}${COLOR_RESET}
+"
+
+LATEST_RELEASE=$(_get_latest_release)
```
lunika, line 1045: What happen if github is down ? Does the script will be usable or not because of this failure ?

bin/arnold
jmaupetit: This is an abstraction layer in case we decide to change our mind and bundle it with Arnold for example. I find it convenient to define all dependencies in the same code block. But of course, this can be changed.

WDYT @madmatah?

lunika: Ok but at the end of the script you added a link to the curl website to install it. What will you do if you change the value of this variable ?

madmatah: I think it's a good practice and consistent with other tools used in this script.

BUT I have a problem with the way it's used line 971 to 975, that limit the possible values here.

jmaupetit: > BUT I have a problem with the way it's used line 971 to 975, that limit the possible values here.

True. Should I remove this check then? What do you suggest?

bin/arnold
jmaupetit: Shellcheck recommends to do so. See https://github.com/koalaman/shellcheck/wiki/SC2155

bin/arnold
lunika: removing the `CURL` variable

bin/arnold
jmaupetit: Dany.

bin/arnold
jmaupetit: I tried to mimic python logging module logic here. See: https://docs.python.org/3/howto/logging.html#logging-levels

bin/arnold
jmaupetit: Good point! I should fix that.

bin/arnold
madmatah: Same here, I would remove the variable and the `eval` L263.

bin/arnold
jmaupetit: Done

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+
+  if [[ -z ${LATEST_VERSION} ]]; then
+    LATEST_VERSION=$(_get_latest_version)
+  fi
+
+  release_url=$(
+    echo "${LATEST_RELEASE}" | \
+    ${JQ} .[0].tarball_url
+  )
+  upgraded_bin="/tmp/arnold-${LATEST_VERSION}"
```
madmatah: I suggest to use `mktemp` to follow best practices in temporary files creation (using $TMPDIR or fallback to /tmp).

```suggestion
  upgraded_bin=$(mktemp -q "arnold-${LATEST_VERSION}-XXXX")
```

Or simply : 

```suggestion
  upgraded_bin=$(mktemp -q)
```

bin/arnold
```diff
@@ -0,0 +1,1009 @@
...
+    log error "You are already up-to-date"
+    exit 10
+  fi
+
+  log debug "Current arnold script path: ${SCRIPT_PATH}"
+
+  upgraded_bin=$(_download_latest_release)
+
+  chmod +x "${upgraded_bin}"
+  cp -i "${upgraded_bin}" "${SCRIPT_PATH}"
```
madmatah: 1/ Why `cp` and not `mv` ? No need to clean-up, then ;-)

2/ The `-i` arguments prevent from running an upgrade without a TTY. Is this the desired behaviour?
Maybe add a `--force` option to your upgrade command without interactive confirmation ?

bin/arnold
jmaupetit: Done both. Thanks!

jmaupetit:
Aaaaand fixed. Ping @lunika && @madmatah :pray:

bin/arnold
```diff
@@ -0,0 +1,1022 @@
...
+    # be defined
+    K8S_AUTH_API_KEY="$(${KUBECTL} whoami -t)"
+    export K8S_AUTH_API_KEY
+    K8S_AUTH_HOST="$(_get_k8s_auth_host)"
+    export K8S_AUTH_HOST
+}
+
+# _ensure_k8s_project_is_set: ensure that k8s's target project has
+# been set via the ARNOLD_CUSTOMER & ARNOLD_ENVIRONMENT environment variables
+function _ensure_k8s_project_is_set() {
```
madmatah: Minor suggestion to use k8s vocabulary : 

```suggestion
function _ensure_k8s_namespace_is_set() {
```

madmatah (approved):
Minor suggestions, but the most important thing to note is :

bin/arnold
jmaupetit: > Why cp and not mv ? No need to clean-up, then ;-)

As it was a temporary file, I though it wasn't worth a cleanup... but ok then.

> The -i arguments prevent from running an upgrade without a TTY. Is this the desired behaviour?

Yes a kind of lazy confirmation step.

> Maybe add a --force option to your upgrade command without interactive confirmation ?

Good idea.

bin/arnold
jmaupetit: Of course!

bin/arnold
lunika: First time I see this... I'm not convinced it's a good thing but why not.

lunika (approved):

PR #592, opened 2021-01-18
✨(cli) improve container volumes management

jmaupetit:
### `env_type`

When we need to override default env_type vars definition, it'd usefull to redefine them in the local arnold project. The default expected behavior is to mount the projetc's `groups_vars/env_type` directory if it has been defined locally.

### volumes

When working with Arnold, we might want to add specific resources or override a container directory (e.g. `apps/`) in the execution context. To fill the gap, we adding an `--extra-volume` option would be usefull

### k8s_domain

Setting the K8S_DOMAIN context by using an option instead of an environment variable would be more user friendly.

## Proposal

- [x] add `groups_vars/env_type` directory override support
- [x] add `--extra-volume` option support
- [x] add `--k8s-domain` option support

### DIFF
--- a/bin/arnold
+++ b/bin/arnold
@@ -69,8 +69,10 @@ declare -A LOG_LEVEL_COLORS=(
 # Directories
 declare -r RELATIVE_COMMON_DIR="group_vars/common"
 declare -r RELATIVE_CUSTOMER_DIR="group_vars/customer"
+declare -r RELATIVE_ENV_TYPE_DIR="group_vars/env_type"
 declare -r COMMON_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_COMMON_DIR}"
 declare -r CUSTOMER_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_CUSTOMER_DIR}"
+declare -r ENV_TYPE_DIR="${ARNOLD_PROJECT_PATH}/${RELATIVE_ENV_TYPE_DIR}"
 
 # Files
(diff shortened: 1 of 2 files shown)


bin/arnold
```diff
@@ -487,6 +495,18 @@ function _docker_run() {
       -v ${COMMON_DIR}:/app/${RELATIVE_COMMON_DIR} \
       -v ${CUSTOMER_DIR}:/app/${RELATIVE_CUSTOMER_DIR}"
 
+    # Override default env_types if the local env_type dir exists
+    if [[ -e ${ENV_TYPE_DIR} ]]; then
+      extras_volumes="${extras_volumes} \
```
sampaccoud: The "s" in "extra" is confusing and I wondered why there is another ${ARNOLD_EXTRA_VOLUMES} variable with a similar name but holding something different.

From what I understand, this variable would be better named: "volumes" or "mounted_volumes". They are not the extras, they are all the volumes to mount.

sampaccoud (approved):

bin/arnold
jmaupetit: In fact the only required volume is the `${HOME}/.kube` volume. I decided to consider all other volumes as "extra volumes". You are right about the typo, should be `extra_volumes`. The `ARNOLD_EXTRA_VOLUMES` var stores volumes passed to the CLI option `--extra-volume`.

sampaccoud: makes sense thanks for the clarification.

PR #595, opened 2021-01-19
🔖(major) bump release to 6.0.0

jmaupetit:
### Changed

- CLI: mount local `env_type` directory if it exists
- CLI: the new `--extra-volume` option allows to mount an extra volume in the arnold container
- CLI: implement `--k8s-domain` option support

### Fixed

- Fix typos in setting names: `withelist` vs `whitelist` (BC)


jmaupetit:
Closed in favor of #596 (we had a chat about this: making a minor release is a better strategy than respecting semver _stricto sensu_).

PR #604, opened 2021-02-11
🔨(k3d) add support for RWX volumes in k3d

madmatah:
## Purpose

Some Arnold apps require `PersistentVolumeClaim` with _ReadWriteMany_ (RWX) acces mode. But out of the box, `k3d` is bundled with a `local-path` StorageClass and a dynamic provisionner that can only provision `PersistentVolumes` with _ReadWriteOnce_ access mode.

As a result, our volume claims cannot be satisfied.

## Proposal

To solve that, since it concerns only the dev and CI environments, the `bin/init-cluster` script pre-provision a bunch of PersistentVolume each time we start the `k3d` cluster. These pre-provisionned PVs are only virtual and do not consume or reserve any resource until they are bound to a PVC.

The data stored on these PVs are written in a docker volume.

## Alternative solution

If you don't like this proposal, as an alternative, we can install a dynamic provisionner like [this one](https://github.com/rimusz/hostpath-provisioner).
The project is based on this [example provisioner](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/tree/master/examples/hostpath-provisioner) and could be used for our needs. But the project seems to be unmaintained. So, do we want that ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,8 +18,9 @@ env.d/*
 .pytest_cache
 .pylint.d
 
-# OpenShift
+# local storage
 openshift.local.clusterup
+k3d-storage
 
 # Ignore local customers except eugene
 group_vars/customer/*
--- a/bin/init-cluster
+++ b/bin/init-cluster
@@ -11,8 +11,37 @@ K3D_CLUSTER_NAME="${1:-arnold}"
 K3D_BIND_HOST_PORT_HTTP="${K3D_BIND_HOST_PORT_HTTP:-8080}"
 K3D_BIND_HOST_PORT_HTTPS="${K3D_BIND_HOST_PORT_HTTPS:-8443}"
 
+# Arnold project root directory
+declare ARNOLD_ROOT
+ARNOLD_ROOT="$(dirname "$(dirname "$(readlink -f "$0")")")"
+
+# Path to the directory that will store the persistent volume inside k3d
+declare -r PERSISTENT_VOLUME_PATH="${ARNOLD_ROOT}/k3d-storage"
+
+# Dependencies to download
 declare -r INGRESS_NGINX_DEPLOYMENT="https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v0.43.0/deploy/static/provider/cloud/deploy.yaml"
 declare -r INGRESS_NGINX_DEPLOYMENT_SHA256="03d80b59660a1818dea0f1b3482aa59cfae2bebcc755b150920ddfa26017adc3"
+
+# PersistentVolume provisionning.
+#
+# Some Arnold apps require PersistentVolume with
+# ReadWriteMany (RWX) acces mode.
+#
+# k3d is bundled with a "local-path" StorageClass using a dynamic
+# provisionner. But it can only provision PersistentVolumes with
+# ReadWriteOnce access mode.
+#
+# Since this single node k3d cluster is only used for development
+# and CI purpose, we pre-provision a certain amount of
+# PersistentVolumes that will be mounted in subdirectories
+# of the docker volume PERSISTENT_VOLUME_PATH.
+#
+# Each time the k3d cluster starts, we ensure that there is
+# at least MINIMUM_AVAILABLE_RWX_VOLUME PVs available.
+#
+# These pre-provisionned PVs are virtual, no resources are reserved
+# for them. And no disk space is consumed until they are bound to a PVC.
+MINIMUM_AVAILABLE_RWX_VOLUME=${MINIMUM_AVAILABLE_RWX_VOLUME:-100}
 
 
 if [[ -z "${K3D_CLUSTER_NAME}" ]] ; then
@@ -22,10 +51,17 @@ fi
 
 
 if ! k3d cluster list "${K3D_CLUSTER_NAME}" &> /dev/null ; then
+
+  # Ensure that the persistent volume storage directory exists on the host
+  if [[ ! -d "${PERSISTENT_VOLUME_PATH}" ]] ; then
+    mkdir -p "${PERSISTENT_VOLUME_PATH}"
+  fi
+
   # Initialize the cluster without the default ingress controller (traefik)
   k3d cluster create "${K3D_CLUSTER_NAME}" \
     -p "${K3D_BIND_HOST_PORT_HTTP}:80@loadbalancer" \
     -p "${K3D_BIND_HOST_PORT_HTTPS}:443@loadbalancer" \
+    -v "${PERSISTENT_VOLUME_PATH}:/data/pv" \
     --k3s-server-arg '--no-deploy=traefik'
 else
   k3d cluster start "${K3D_CLUSTER_NAME}"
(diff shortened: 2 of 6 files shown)


bin/init-cluster
```diff
@@ -22,10 +52,17 @@ fi
 
 
 if ! k3d cluster list "${K3D_CLUSTER_NAME}" &> /dev/null ; then
+
+  # Ensure that the longhorn directory exists on the host
```
jmaupetit: longhorn?

jmaupetit (commented):
I'm not sure to understand what you are achieving here. Can you help me figure this out:

1. How will we use those PVs in development? Should we have development specific configuration or will there be some k8s-black magic I don't get here?
2. What happen when all PVs have been reached?

madmatah:
> I'm not sure to understand what you are achieving here. Can you help me figure this out:
> 
> 1. How will we use those PVs in development? Should we have development specific configuration or will there be some k8s-black magic I don't get here?

There is no black magic here :grin: 
This is just manual provisionning of [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)  : 

- The administrator declares a set of `PersistentVolume`s with their specifications (capacity, access mode)
- Ther are created with the `Available` status.
- When a `PersistentVolumeClaim` is created, if the requirements matches with one of the Available PV, it is bound to it. And the status of the PV changes to «Bound».

The only «magic» thing here is the storageClass.
A k8s cluster can handle multiple storageClass, and they must match between a PV and a PVC to be bound.

Out of the box, k3d is configured with the **default** storageClass `local-path`.
The PVs created by the `init-cluster` script are also defined with the storageClass `local-path` (even if they are not managed by the dynamic provisionner bundled by k3d).

Since our apps request volumes with no storageClass, the default one is used.
And since our apps request volumes with RWX access mode, only PVs that we manually provisionned will match.
If an app request a volume with RWO access mode, it will be handled by the dynamic volume provisionner.

Is it clearer now ?

We are used to dynamic (or on demand) provisionning of PV with advanced storageClasses, but k8s static provisionning can still be helpful!


> 2. What happen when all PVs have been reached?

When all PVs are bound, all future PVC will be pending.
That's why, at each startup, the init-cluster ensure that at least `MINIMUM_AVAILABLE_RWX_VOLUME` (50) PVs exist with the `Available` status. If not, new PVs are created to reach this number.

bin/init-cluster
madmatah: Oopsie, this is a residue of a test I made with a storageClass named Longhorn, that doesn't work with k3d ;)

lunika:
Clever solution !
What can be the side effects we don't want ? Race condition on accessing a file ?

madmatah:
> Clever solution !
> What can be the side effects we don't want ? Race condition on accessing a file ?

@lunika Thanks!

The main drawback I just discovered is specific to the usage of [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) : 

> The files or directories created on the underlying hosts are only writable by root. You either need to run your process as root in a privileged Container or modify the file permissions on the host to be able to write to a hostPath volume

But all solutions I see to solve this problem are based on `hostPath` 😕

The other drawbacks I see here are that : 
 - This approach is limited to a single node k8s cluster
 - If the number of PVs is reached, it requires to execute `make cluster` to provision more PVs.

I think these are acceptable for a dev / CI environment.

I don't see race condition effects specific to this solution that we should be aware of.

madmatah:
Additional information I just found:

I was wondering : How is storage handled by Openshift when we launch a simple `oc cluster up` ?
The answer is fun: 100 PVs based on hostPath are created during cluster initialization.

They look like this :

```
apiVersion: v1
kind: PersistentVolume
metadata:
  creationTimestamp: 2020-10-20T15:56:58Z
  finalizers:
  - kubernetes.io/pv-protection
  labels:
    volume: pv0070
  name: pv0070
  resourceVersion: "2205"
  selfLink: /api/v1/persistentvolumes/pv0070
  uid: e2a7945e-12ec-11eb-801a-18d6c71f59d1
spec:
  accessModes:
  - ReadWriteOnce
  - ReadWriteMany
  - ReadOnlyMany
  capacity:
    storage: 100Gi
  hostPath:
    path: /home/example/openshift.local.clusterup/openshift.local.pv/pv0070
    type: ""
  persistentVolumeReclaimPolicy: Recycle
status:
  phase: Available
```

So I think we'll have an isofunctional storage solution :wink:

lunika:
> Additional information I just found:
> 
> I was wondering : How is storage handled by Openshift when we launch a simple `oc cluster up` ?
> The answer is fun: 100 PVs based on hostPath are created during cluster initialization.
> 
> They look like this :
> 
> ```
> apiVersion: v1
> kind: PersistentVolume
> metadata:
>   creationTimestamp: 2020-10-20T15:56:58Z
>   finalizers:
>   - kubernetes.io/pv-protection
>   labels:
>     volume: pv0070
>   name: pv0070
>   resourceVersion: "2205"
>   selfLink: /api/v1/persistentvolumes/pv0070
>   uid: e2a7945e-12ec-11eb-801a-18d6c71f59d1
> spec:
>   accessModes:
>   - ReadWriteOnce
>   - ReadWriteMany
>   - ReadOnlyMany
>   capacity:
>     storage: 100Gi
>   hostPath:
>     path: /home/example/openshift.local.clusterup/openshift.local.pv/pv0070
>     type: ""
>   persistentVolumeReclaimPolicy: Recycle
> status:
>   phase: Available
> ```
> 
> So I think we'll have an isofunctional storage solution wink

Great ! You should maybe set the default to 100 and not 50 in this case ?

madmatah:
> Great ! You should maybe set the default to 100 and not 50 in this case ?

Good suggestion, Done!
I set the default to 100 in the `init-cluster` script.

But I reduced it in CircleCI job to 15 to speed-up cluster bootstrap time.
Since each job has its own k3d cluster and only 1 app is deployed per job, I think it's enough. WDYT ?

madmatah:
I added a `ReadWriteMany` volume to the `hello` app and a job to ensure that an unprivileged container can write into the volume.

I used the same strategy that Openshift for permission management : 
 - PV directories are pre-created with `root:root` ownership and perm `770`.
   We don't want to delegate the creation of these directories to kubernetes, because it will create them with insufficient permissions.
- We consider that containers are running with the securityContext param `runAsGroup: 0` (which seems to be the default behavior).

.gitignore
```diff
@@ -18,8 +18,9 @@ env.d/*
 .pytest_cache
 .pylint.d
 
-# OpenShift
+# OpenShift / K8S
```
jmaupetit:
```suggestion
# local storage
```

bin/init-cluster
```diff
@@ -11,9 +11,39 @@ K3D_CLUSTER_NAME="${1:-arnold}"
...
+# Dependencies to download
 declare -r INGRESS_NGINX_DEPLOYMENT="https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v0.43.0/deploy/static/provider/cloud/deploy.yaml"
 declare -r INGRESS_NGINX_DEPLOYMENT_SHA256="03d80b59660a1818dea0f1b3482aa59cfae2bebcc755b150920ddfa26017adc3"
 
+# PersistentVolume provisionning.
+#
+# Some Arnold apps require PersistentVolume with
+# ReadWriteMany (RWX) acces mode.
+#
+# k3d is bundled with a "local-path" StorageClass with a dynamic
```
jmaupetit:
```suggestion
# k3d is bundled with a "local-path" StorageClass using a dynamic
```

bin/init-cluster
```diff
@@ -11,9 +11,39 @@ K3D_CLUSTER_NAME="${1:-arnold}"
...
+# and CI purpose, we pre-provision a certain amount of
+# PersistentVolumes that will be mounted in subdirectories
+# of the docker volume PERSISTENT_VOLUME_PATH.
+#
+# Each time the k3d cluster starts, we ensure that there is
+# at least MINIMUM_AVAILABLE_RWX_VOLUME PVs available.
+#
+# These pre-provisionned PVs are virtual, no resources are reserved
+# for them. And no disk space is consumed until they are bound to a PVC.
+
```
jmaupetit:
```suggestion
# for them. And no disk space is consumed until they are bound to a PVC.
```

bin/init-cluster
```diff
@@ -34,6 +71,49 @@ fi
...
+# Pre-provision PersistentVolumes with RWX accessModes
+
+# Count the number of PVs with RWX accessMode and with the status Available
+AVAILABLE_PV_COUNT=$(kubectl get pv \
+  -l type=arnold-rwx \
+  -o jsonpath='{range .items[?(@.status.phase=="Available")]}{.metadata.name}{"\n"}{end}' \
+  | wc -l)
+
+# Ensure that there is at least MINIMUM_AVAILABLE_RWX_VOLUME PVs available
```
jmaupetit:
```suggestion
# Ensure that there are at least MINIMUM_AVAILABLE_RWX_VOLUME PVs available
```

bin/init-cluster
```diff
@@ -34,6 +71,49 @@ fi
...
+if [[ $AVAILABLE_PV_COUNT -lt $MINIMUM_AVAILABLE_RWX_VOLUME ]] ; then
+  PV_TO_PROVISION=$(( MINIMUM_AVAILABLE_RWX_VOLUME -  AVAILABLE_PV_COUNT ))
+  echo "Provisionning ${PV_TO_PROVISION} PersistentVolumes with RWX accessMode"
+  for _ in $(seq 1 $PV_TO_PROVISION) ; do
+    # Generate an unique identifier for the PV
+    PV_ID=$(uuidgen)
+    # Pre-create the directory with enough permissions so that an unprivileged container
+    # can write into it. For that, we give write permission to the group 0 and assume that
+    # containers are launched with `runAsGroup: 0`.
+    # We do this because if the directory were created by the hostPath handler, only privileged
```
jmaupetit:
```suggestion
    # We do this because if the directory was created by the hostPath handler, only privileged
```

jmaupetit (approved):
Wonderful :muscle:

PR #607, opened 2021-02-16
✨(core) allow to specify uid of running containers

madmatah:
## Purpose

By default, kubernetes will run containers with the USER specified in
the docker image. And this USER can be root. We want to run containers
as non-root user, like Openshift was doing by default.

Kubernetes allow to specify a SecurityContext for resources that will
create containers (e.g. Pods, Deployments, Jobs...). SecurityContext
has a RunAsUser option which allows to run containers with the
specified UID.

## Proposal

A new ansible variable `container_uid` has been introduced to specify
the UID that will be used to run containers. We recommend to specify a
unique UID per namespace.

The `hello` app has been updated to use it. And this will have to be
generalized to all arnold apps.

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -163,3 +163,18 @@ network_policy_rule_ingress_controller:
       matchLabels:
         "app.kubernetes.io/component": controller
         "app.kubernetes.io/name": ingress-nginx
+
+
+# UID to use for running containers on k8s.
+#
+# It will be used in k8s templates to specify the SecurityContext.RunAsUser
+# field, to force pods to run with this specific UID.
+#
+# It prevents from running a container as root.
+#
+# We recommend that you specify a unique UID per customer/environment.
+container_uid: 10000
+
+# GID to use for running containers on k8s.
+# This value should be set to 0 (root).
+container_gid: 0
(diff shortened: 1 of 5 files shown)


group_vars/all/main.yml
```diff
@@ -163,3 +163,18 @@ network_policy_rule_ingress_controller:
...
+
+
+# UID to use for running containers on k8s.
+#
+# It will be used in k8s templates to specify the SecurityContext.RunAsUser
+# field, to force pods to run with this specific UID.
+#
+# It prevents from running a container as root.
+#
+# We recommend that you specify an unique UID per customer or environment.
```
```suggestion
# We recommend that you specify a unique UID per customer or customer/environment.
```

jmaupetit (approved):
Smart move!

group_vars/all/main.yml
sampaccoud: Why not emulate openshift by running on random uids by default?

lunika:
Do we have to add this section on all apps ?

madmatah:
> Do we have to add this section on all apps ?

Yes 😕

group_vars/all/main.yml
madmatah: In fact, Openshift generates a random UID range per namespace and allows us to use all UIDs of this range.
But according to [their documentation](https://www.openshift.com/blog/a-guide-to-openshift-and-uids) :

> When a Pod is deployed into the namespace, by default, OpenShift will use the first UID and first GID from this range to run the Pod.

And since we don't specify manually UIDs for our pods, in fact, we are always using the same UID per namespace.
And the solution proposed in this PR is the same, IMO.

lunika (approved):

PR #582, opened 2020-12-11
📝(project) update documentation using the CLI

jmaupetit:
## Proposal

Now that we have a CLI to use arnold, our documentation needed a solid revamp to reflect recent changes.

The documentation no longer focuses on `edxapp` but the tool itself, with a clear distinction between tool users and developers.

Depends on #576

PR #609, opened 2021-02-17
🚚(core) remove the concept of project in arnold

madmatah:
## Purpose

In k8s, the `Project` object type doesn't exist, it is called a `Namespace`.

## Proposal

All occurrences of project have been replaced with `namespace` to avoid confusion.


lunika (commented):
There are references to `project` in the documentation. Did you plan to work on the documentation later ? Or to work on each part in the associated PR ?

madmatah:
> There are references to `project` in the documentation. Did you plan to work on the documentation later ? Or to work on each part in the associated PR ?

@lunika  Yes, I deliberately left the documentation aside with the goal of rewriting it later.
This is a big work, some parts are deprecated, some other parts have already been rewritten in #582. So I think it will be more efficient to rewrite it in a dedicated PR.

bin/arnold
```diff
@@ -182,34 +182,34 @@ OPTIONS:
 
 COMMANDS:
 
-> Prepare current project:
+> Prepare current environment:
```
jmaupetit: I think there might be a confusion between what is called here an `environment` and the `env_type`. Does an `environment` means a customer/env_type combination to you?

If yes, I think we should still call it a namespace as it corresponds to a customer/env_type couple.

madmatah: Yes, it means a customer/env_type combination to me.
Ok, I'll do what you recommend.

madmatah: Done :heavy_check_mark: 
I also updated the commit message body.

lunika (approved):

PR #610, opened 2021-02-18
🔧(redis) update redis to be compatible with k8s

madmatah:
## Purpose

We want to deploy redis on a k8s cluster instead of Openshift.

## Proposal

- [x] Update templates
- [x] Remove `initialDelaySeconds` in readinessProbe to speedup startup time

### DIFF
--- a/apps/redis/templates/services/app/deploy.yml.j2
+++ b/apps/redis/templates/services/app/deploy.yml.j2
@@ -56,7 +56,6 @@ spec:
                 - "/bin/bash"
                 - "-c"
                 - "redis-cli ping && touch /data/healthcheck"
-            initialDelaySeconds: 60
             periodSeconds: 10
           volumeMounts:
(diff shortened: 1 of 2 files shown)


apps/redis/templates/services/app/deploy.yml.j2
```diff
@@ -50,7 +56,6 @@ spec:
                 - "/bin/bash"
                 - "-c"
                 - "redis-cli ping && touch /data/healthcheck"
-            initialDelaySeconds: 60
             periodSeconds: 10
           volumeMounts:
             - mountPath: /data
```
jmaupetit, line 61: We don't specify a container user here?

madmatah: Because I forgot :grimacing: , thanks!

jmaupetit (approved):

PR #615, opened 2021-02-18
🔧(mailcatcher) update mailcatcher to be compatible with k8s

madmatah:
## Purpose

We want to deploy mailcatcher on a k8s cluster instead of Openshift.

## Proposal

- [x] Fix mailcatcher configuration templates

## How to test

- `make cluster`
- `ANSIBLE_VAULT_PASSWORD=arnold bin/arnold -a mailcatcher -c eugene -d -e development bootstrap`
- `echo "https://next.mailcatcher.development-eugene.$(hostname -I | awk '{print $1}').nip.io:8443/development/"`


apps/mailcatcher/templates/services/app/deploy.yml.j2
```diff
@@ -9,12 +9,16 @@ metadata:
   namespace: "{{ namespace_name }}"
 spec:
   replicas: 1  # number of pods we want
+  selector:
+    matchLabels:
+      app: mailcatcher
+      service: app
+      deploymentconfig: "mailcatcher-app"
```
lunika: We still use this label ? I think you should rename it in `deployment` here and line 22

madmatah: Good catch, thanks!
Yes, this label is used by the service.

lunika (approved):

PR #622, opened 2021-02-23
🐛(scripts) use the right k8s context in init-cluster

madmatah:
## Purpose

The `bin/init-cluster` script executes `kubectl` commands without
specifying the configuration file related to the k3d cluster. So, the
current kubectl context is used instead. As a result, the commands
related to PV pre-provisionning and ingress-nginx controller could be
executed on another cluster.

## Proposal

- [x] Add `--kubeconfig` parameter to all `kubectl` commands executed

### DIFF
--- a/bin/init-cluster
+++ b/bin/init-cluster
@@ -74,7 +74,7 @@ echo "K3d cluster configuration exported to ${KUBECONFIG}"
 # Pre-provision PersistentVolumes with RWX accessModes
 
 # Count the number of PVs with RWX accessMode and with the status Available
-AVAILABLE_PV_COUNT=$(kubectl get pv \
+AVAILABLE_PV_COUNT=$(kubectl --kubeconfig="${KUBECONFIG}" get pv \
   -l type=arnold-rwx \
   -o jsonpath='{range .items[?(@.status.phase=="Available")]}{.metadata.name}{"\n"}{end}' \
   | wc -l)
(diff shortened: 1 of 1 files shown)


bin/init-cluster
```diff
@@ -74,7 +74,7 @@ echo "K3d cluster configuration exported to ${KUBECONFIG}"
 # Pre-provision PersistentVolumes with RWX accessModes
 
 # Count the number of PVs with RWX accessMode and with the status Available
-AVAILABLE_PV_COUNT=$(kubectl get pv \
+AVAILABLE_PV_COUNT=$(kubectl --kubeconfig="${KUBECONFIG}" get pv \
```
jmaupetit, line 77: Where/when is defined this environment variable?

madmatah: @jmaupetit  Line 70 : 
```
KUBECONFIG=$(k3d kubeconfig write "${K3D_CLUSTER_NAME}")
```

jmaupetit: Okay, I missed it :sweat_smile: ... sorry for the noise.

bin/init-cluster
```diff
@@ -74,7 +74,7 @@ echo "K3d cluster configuration exported to ${KUBECONFIG}"
 # Pre-provision PersistentVolumes with RWX accessModes
```
jmaupetit, line 74: In your commit message: `s/PV pre-previsionning/PV pre-provisionning/`

jmaupetit (approved):

bin/init-cluster
madmatah: Thanks!

PR #623, opened 2021-02-23
✨(core) define a default storage class with ansible variable

madmatah:
## Purpose

A default `StorageClass` is not necessarily defined on a k8s cluster.
When this is the case, the `PersistentVolumeClaim` created by arnold are
stuck in the _Pending_ state.

(This PR depends on #622) 

## Proposal

- [x] Introduce new ansible variables `default_storage_class_rwx` and `default_storage_class_rwo` to define the Storage Class to use by default
- [x] Update apps volume templates to use these
- [x] Change the storage class name of the RWX PersistentVolumes that we pre-provision in the `bin/init-cluster` script, so they are not related to the default storage class anymore. This will allow the CI to validate that everything is working well.

### DIFF
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -178,3 +178,13 @@ container_uid: 10000
 # GID to use for running containers on k8s.
 # This value should be set to 0 (root).
 container_gid: 0
+
+
+# Storage class
+# A default storage class is not necessarily defined in a k8s cluster. The following
+# variables defines the storage class to use for our PVC, depending on the accessMode
+# we are requesting (RWO = ReadWriteOnce, RWX = ReadWriteMany).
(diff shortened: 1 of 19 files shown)


group_vars/all/main.yml
```diff
@@ -178,3 +178,14 @@ container_uid: 10000
...
 container_gid: 0
+
+
+# Storage class
+# A default storage class is not necessarily defined in a k8s cluster. The following
+# variables defines the storage class to use for our PVC, depending on the accessMode
+# we are requesting (RWO = ReadWriteOnce, RWX = ReadWriteMany).
+# If an arnold app requires a specific storageClass, it should be handled in the app
+# directly.
+
```
```suggestion
# directly.
```

jmaupetit (approved):

PR #625, opened 2021-02-25
Fix/misc k8s fixes

madmatah:
## Purpose

This PR is intended to fix 5 issues : 

1. The `postgresql` Deployment template in `Ashley` causes an error in non-trashable environment type.
2. There is a confict in the secret name used to store TLS certificates in blue/green applications
3. The annotation used to declare the cert-manager `Issuer` to use in invalid in all Ingresse templates
4. There is warning due to the usage of a deprecated `apiVersion` in Ingresses templates
5. The `postgresql` Deployment template in `Richie` causes an error in non-trashable environment type.

## Proposal

1. [x] Fix Deployment template for ashley/postgresql
2. [x] Use the prefix in the secret name to have a distinct secret per Ingress
3. [x] Use `cert-manager.io/issuer` annotation instead of `cert-manager.io/cluster-issuer`
4. Upgrade all Ingresses :
   - [x] Change apiVersion `extensions/v1beta1` :arrow_right:  `networking.k8s.io/v1`
   - [x] Add [pathType](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) argument that is now required 
   - [x] Update the backend definition according to the new spec
5. [x] Fix Deployment template for richie/postgresql

### DIFF
--- a/apps/ashley/templates/services/nginx/ingress.yml.j2
+++ b/apps/ashley/templates/services/nginx/ingress.yml.j2
@@ -25,4 +25,4 @@ spec:
   tls:
   - hosts:
     - "{{ ashley_host | blue_green_host(prefix) }}"
-    secretName: ashley-app-tls
+    secretName: "ashley-app-tls-{{ prefix }}"
--- a/apps/ashley/templates/services/nginx/ingress.yml.j2
+++ b/apps/ashley/templates/services/nginx/ingress.yml.j2
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   namespace: "{{ namespace_name }}"
@@ -19,9 +19,12 @@ spec:
     http:
       paths:
       - backend:
-          serviceName: "ashley-nginx-{{ prefix }}"
-          servicePort: {{ ashley_nginx_port }}
+          service:
+            name: "ashley-nginx-{{ prefix }}"
+            port:
+              number: {{ ashley_nginx_port }}
         path: /
+        pathType: Prefix
   tls:
   - hosts:
     - "{{ ashley_host | blue_green_host(prefix) }}"
--- a/apps/ashley/templates/services/nginx/ingress.yml.j2
+++ b/apps/ashley/templates/services/nginx/ingress.yml.j2
@@ -12,7 +12,7 @@ metadata:
     route_prefix: "{{ prefix }}"
     route_target_service: "app"
   annotations:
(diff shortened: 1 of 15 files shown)


apps/ashley/templates/services/postgresql/deploy.yml.j2
```diff
@@ -44,7 +44,7 @@ spec:
           envFrom:
             - secretRef:
                 name: "{{ ashley_database_secret_name }}"
-{% endif%}
       securityContext:
         runAsUser: {{ container_uid }}
         runAsGroup: {{ container_gid }}
+{% endif%}
```
jmaupetit, line 50: Is this the only impacted template?

apps/ashley/templates/services/nginx/ingress.yml.j2
```diff
@@ -12,7 +12,7 @@ metadata:
     route_prefix: "{{ prefix }}"
     route_target_service: "app"
   annotations:
-    cert-manager.io/cluster-issuer: "{{ acme_issuer_name }}"
+    cert-manager.io/issuer: "{{ acme_issuer_name }}"
```
jmaupetit, line 15: Should it be configurable?

apps/ashley/templates/services/postgresql/deploy.yml.j2
madmatah: Yes

apps/ashley/templates/services/nginx/ingress.yml.j2
madmatah: Since arnold explicitly creates an Issuer per namespace during the project bootstrap, I don't think this is necessary in the current state of the project.

Should it be configurable per project, or across all arnold apps ?
It might be better to develop this in another PR, if we want to do it.

apps/ashley/templates/services/postgresql/deploy.yml.j2
madmatah: Uh, oh. I was sure I checked everything.
But I double checked and apparently my memories are not reliable :sweat_smile: 
I'll fix this.

madmatah: Done in 3d294c4b469b6997bb02bc441159dedbcc56b5ca

apps/ashley/templates/services/nginx/ingress.yml.j2
jmaupetit: ok then

jmaupetit (approved):

PR #628, opened 2021-03-02
✨ allow to switch ACME environment easily

madmatah:
## Purpose

We want to be able to switch the ACME environment used for TLS certificates generation in a namespace
(e.g. To switch from the `staging` to the `live` ACME environment).

This process is difficult in the current arnold version, because it requires to execute manual operations to delete entities automatically managed by `cert-manager` to avoid caching issues.

This is due to the fact that the name of the "Issuer" is the same regardless of the environment used.
And also, the secret name used to store the certificate is the same regardless of the environment used.

## Proposal

- [x] add a `acme` command to arnold CLI to create/update the Issuer of a namespace
- [x] fix error when the Issuer already exist in a namespace
- [x] Update TLS secretName used in arnold apps' ingresses to include the acme environment name
- [x] Update the Issuer name so it contains the acme environment name
- [x] Fix redirect ingress template that dit not follow the latest fixes and improvements.

## Example scenario : 

1. Bootstrap a project with `acme_env: "staging"`  (`arnold bootstrap`)
2. Change the `acme_env` variable to `"live"`
3. Execute `arnold acme` to update the Issuer
4. Execute `arnold ingresses` to update the ingresses
5. :tada:


lunika (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- CLI: add `acme` command to create or update the namespace Issuer
```
jmaupetit, line 13:
```suggestion
- CLI: add `acme` command to create or update the namespace SSL certificate Issuer
```

apps/ashley/templates/services/nginx/ingress.yml.j2
```diff
@@ -28,4 +28,4 @@ spec:
   tls:
   - hosts:
     - "{{ ashley_host | blue_green_host(prefix) }}"
-    secretName: "ashley-app-tls-{{ prefix }}"
+    secretName: "ashley-app-tls-{{ prefix }}-{{acme_env}}"
```
jmaupetit, line 31:
```suggestion
    secretName: "ashley-app-tls-{{ prefix }}-{{ acme_env }}"
```

core_apps/redirect/templates/services/nginx/_redirected_ingress.yml.j2
```diff
@@ -1,25 +1,28 @@
...
+            name: "redirect-nginx"
+            port:
+              number: {{ redirect_nginx_port }}
         path: /
+        pathType: Prefix
   tls:
   - hosts:
     - "{{ redirection_from | default('unknown') }}"
-    secretName: "redirect-app-{{ redirection_from | default('unknown')}}"
+    secretName: "redirect-app-{{ redirection_from | default('unknown')}}-{{acme_env}}"
```
jmaupetit, line 28:
```suggestion
    secretName: "redirect-app-{{ redirection_from | default('unknown') }}-{{ acme_env }}"
```

jmaupetit (commented):
I'm a bit too late to the party, but I would suggest to add spaces to improve readability. This is a convention we try to use in templates.

PR #637, opened 2021-03-04
✨(marsha) add cronjob for management command check_live_idle

lunika:
## Purpose

The management command check_live_idle should be executed once a day.
For this we add a cronjob in marsha app configured to run the management
command every day at 4am

## Proposal

- [x]  add cronjob for management command check_live_idle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 - CLI: add `acme` command to create or update the namespace TLS certificate Issuer
 - Marsha: cronjob running check_harvested management command
+- Marsha: cronjob running check_live_idle management command
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


sampaccoud:
I thought of this subject of multiple cron jobs with different run frequency.
Couldn't we run only one job with the highest frequency and take care in the app that the job that does not need such high frequency works as well :man_shrugging:  ? 
The idea here is to limit the complexity on the ops side. Avoid leaking application problems to the ops.

lunika:
What you suggest is to merge all management command in only one and then deal with the frequency in the code itself ?

sampaccoud:
> What you suggest is to merge all management command in only one and then deal with the frequency in the code itself ?

Yep. Because we know many more are coming and we don't like to multiply the number of jobs and the complexity of ops?

madmatah (approved):
I approve since this feature is already merged on `master` and the goal of `migration-to-k8s` branch is to be isofunctional.

But I think it would be interesting to explore the idea of having a single cronjob to rule them all, in another PR on marsha.

PR #639, opened 2021-03-08
🐛(cli) autodetect if we are in a terminal

madmatah:
## Purpose

We want to use arnold CLI on CI environements that don't run
necessarily inside a terminal. In these case, the docker run command
fail with the following error: `the input device is not a TTY`,
because of the `-t` option.

## Proposal

Autodetect if we are within a terminal and add the `-t` option to docker run only in this case.

### DIFF
--- a/bin/arnold
+++ b/bin/arnold
@@ -604,8 +604,15 @@ function _docker_run() {
       fi
     fi
 
+    if tty > /dev/null ; then
+      # If we are in a terminal, add the `-t` option to docker run
+      DOCKER_TERM_OPTIONS="-ti"
+    else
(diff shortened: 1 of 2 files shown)


lunika (approved):

bin/arnold
```diff
@@ -604,8 +604,15 @@ function _docker_run() {
       fi
     fi
 
+    if tty > /dev/null ; then
+      # If we are in a terminal, add the `-t` option to docker run
+      DOCKER_TERM_OPTIONS="-ti"
+    else
+      DOCKER_TERM_OPTIONS="-i"
+    fi
```
jmaupetit, line 612: Nice trick :clap:

PR #641, opened 2021-03-11
🐛(core) trim templates before checking if we should ignore them

madmatah:
## Purpose

We are unable to deploy the forum on non-trashable environment type.
This is due to the fact that the mongodb deployment template contains
an extra-line at the end of file, and is not ignored by the "Ignore
empty rendered templates" task. It results in a fatal error during
deployment. The ansible task has been fixed to trim files and avoid
this kind of error.

## Proposal
- [x] Fix the "Ignore empty rendered template" task
- [x] Remove the extra line in mongodb deployment template

### DIFF
--- a/apps/forum/templates/services/mongodb/deploy.yml.j2
+++ b/apps/forum/templates/services/mongodb/deploy.yml.j2
@@ -45,4 +45,3 @@ spec:
         runAsUser: {{ container_uid }}
         runAsGroup: {{ container_gid }}
 {% endif %}
-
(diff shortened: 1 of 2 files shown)


lunika (approved):

lunika:
It is not possible to trim the content otherwise ? 
Here https://github.com/openfun/arnold/blob/master/tasks/filter_empty_templates_for_app.yml#L15, doing something like: 

```
when: lookup('template', item) | trim | length > 1
```

madmatah:
> It is not possible to trim the content otherwise ?
> Here https://github.com/openfun/arnold/blob/master/tasks/filter_empty_templates_for_app.yml#L15, doing something like:
> 
> ```
> when: lookup('template', item) | trim | length > 1
> ```

You're right, it will be much better.

madmatah:
@lunika Done, thanks for your suggestion!
Sorry there is twice as much code to review.

lunika (approved):

PR #642, opened 2021-03-11
🔧(elasticsearch) update elasticsearch to be compatible with k8s

madmatah:
## Purpose

We want to deploy elasticsearch on a k8s cluster instead of Openshift.

## Proposal

- [x] Add securityContext in elasticsearch configuration templates
- [x] Increase memory limit to be able to run it on dev environment
- [x] Fix issue with secrets when `elasticsearch_security_enabled: false`

## How to test

- `make cluster`
- `ANSIBLE_VAULT_PASSWORD=arnold bin/arnold -a elasticsearch -c eugene -d -e development bootstrap`


madmatah:
The proposed solution to fix the secret issue is not working.
Switching back this PR to WIP...

madmatah:
Here is another proposal to fix the secrets issue secrets when `elasticsearch_security_enabled: false` :
- We create empty secrets in k8s
- A vault is still initialized for the customer (with `bin/arnold create_app_vaults`)

I haven't found a better solution.

jmaupetit (approved):

PR #644, opened 2021-03-16
✨(apps) allow defining resources requests for all pods

madmatah:
## Purpose
We need to improve pod scheduling on our Kubernetes clusters.

This PR is based on the work done in #353 

## Proposal
By setting resources requests without any limit, our pods are in
"Burstable" mode: we can define the minimum resources it requires upon
scheduling, while still allowing it to burst.

- [x] Rebase #353 with `migration-to-k8s`
- [x] Allow to use jinja in app defaults files
- [x]  Add missing apps
  - [x] ashley
  - [x] edxec
  - [x] elasticsearch
  - [x] etherpad
  - [x] flower
  - [x] kibana
  - [x] redis-sentinel
- [x] Add limits in cronjobs
- [x] Set default limits that reflect real usage.

### DIFF
--- a/group_vars/customer/eugene/main.yml
+++ b/group_vars/customer/eugene/main.yml
@@ -13,6 +13,10 @@ nextcloud_database_name: nextcloud
 elasticsearch_java_opts:
   - "-Xms128m"
   - "-Xmx128m"
+elasticsearch_app_resources:
+  requests:
+    cpu: 100m
+    memory: 128Mi
 elasticsearch_image_name: "elasticsearch"
 elasticsearch_image_tag: "7.7.0"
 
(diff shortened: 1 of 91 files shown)


group_vars/customer/eugene/main.yml
```diff
@@ -13,6 +13,10 @@ nextcloud_database_name: nextcloud
 elasticsearch_java_opts:
   - "-Xms128m"
   - "-Xmx128m"
+elasticsearch_app_resources:
```
sampaccoud, line 16: I didn't understand why Eugene needs to override this variable?

sampaccoud (approved):

group_vars/customer/eugene/main.yml
madmatah: Because 512MB (default value) is too much for a dev CI/env, it has been configured to use 128MB instead.

sampaccoud: I get the idea: Eugene is only for dev/CI so let's limit it.

jmaupetit (approved):

madmatah:
Okay everything seems to be fine.
I adjusted resources request values based on running application on a production environment.
It might have to be adjusted again because I didn't have historical data, but it is a good start.

PR #658, opened 2021-04-20
✨(prosody) add new prosody apps

lunika:
## Purpose

Prosody is an XMPP server. This app is added to arnold and this is a non
blue green application. Only the http part (BOSH) of the app is exposed
outside kubernetes cluster. All services are only internal and marsha
will use there service name to access them.

## Proposal

- [x] add prosody app


apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+-- It looks for mod_modulename.lua in the plugins folder, so make sure that exists too.
+-- Documentation for bundled modules can be found at: https://prosody.im/doc/modules
+modules_enabled = {
+{% for module in prosody_general_modules_enabled %}
+  "{{ module }}";
+{% endfor %}
+}
+
+-- These modules are auto-loaded, but should you want
+-- to disable them then uncomment them here:
```
madmatah: The comment is not relevant since there is nothing to uncomment.

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Disable account creation by default, for security
+-- For more information see https://prosody.im/doc/creating_accounts
+allow_registration = {{ prosody_allow_registration | lower }}
+
+-- Force clients to use encrypted connections? This option will
+-- prevent clients from authenticating unless they are using encryption.
+
+c2s_require_encryption = {{ prosody_c2s_require_encryption | lower }}
+allow_unencrypted_plain_auth = {{ prosody_allow_unencrypted_plain_auth | lower }}
```
madmatah:
```suggestion
-- Force clients to use encrypted connections? This option will
-- prevent clients from authenticating unless they are using encryption.
c2s_require_encryption = {{ prosody_c2s_require_encryption | lower }}
allow_unencrypted_plain_auth = {{ prosody_allow_unencrypted_plain_auth | lower }}
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Force servers to use encrypted connections? This option will
+-- prevent servers from authenticating unless they are using encryption.
+
+s2s_require_encryption = {{ prosody_s2s_require_encryption | lower}}
```
madmatah:
```suggestion
-- Force servers to use encrypted connections? This option will
-- prevent servers from authenticating unless they are using encryption.
s2s_require_encryption = {{ prosody_s2s_require_encryption | lower}}
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Force certificate authentication for server-to-server connections?
+
+s2s_secure_auth = {{ prosody_s2s_secure_auth | lower }}
```
madmatah:
```suggestion
-- Force certificate authentication for server-to-server connections?
s2s_secure_auth = {{ prosody_s2s_secure_auth | lower }}
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Some servers have invalid or self-signed certificates. You can list
+-- remote domains here that will not be required to authenticate using
+-- certificates. They will be authenticated using DNS instead, even
+-- when s2s_secure_auth is enabled.
+
+s2s_insecure_domains = { {{ prosody_s2s_insecure_domains | join(',') }} }
```
madmatah:
```suggestion
-- Some servers have invalid or self-signed certificates. You can list
-- remote domains here that will not be required to authenticate using
-- certificates. They will be authenticated using DNS instead, even
-- when s2s_secure_auth is enabled.
s2s_insecure_domains = { {{ prosody_s2s_insecure_domains | join(',') }} }
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Even if you disable s2s_secure_auth, you can still require valid
+-- certificates for some domains by specifying a list here.
+
+s2s_secure_domains = { {{ prosody_s2s_secure_domains | join(',') }} }
+
+-- Required for init scripts and prosodyctl
```
madmatah: I think this comment refers to something that has been removed?

```suggestion
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+
+-- Select the authentication backend to use. The 'internal' providers
+-- use Prosody's configured data storage to store the authentication data.
+
+authentication = "{{ prosody_general_authentication }}"
```
madmatah:
```suggestion
-- Select the authentication backend to use. The 'internal' providers
-- use Prosody's configured data storage to store the authentication data.
authentication = "{{ prosody_general_authentication }}" 
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+-- use Prosody's configured data storage to store the authentication data.
+
+authentication = "{{ prosody_general_authentication }}" 
+
+-- Select the storage backend to use. By default Prosody uses flat files
+-- in its configured data directory, but it also supports more backends
+-- through modules. An "sql" backend is included by default, but requires
+-- additional dependencies. See https://prosody.im/doc/storage for more info.
+
+storage = "{{ prosody_storage }}"
```
madmatah:
```suggestion
-- Select the storage backend to use. By default Prosody uses flat files
-- in its configured data directory, but it also supports more backends
-- through modules. An "sql" backend is included by default, but requires
-- additional dependencies. See https://prosody.im/doc/storage for more info.
storage = "{{ prosody_storage }}"
```

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
```diff
@@ -0,0 +1,166 @@
...
+{% endif %}
+
+
+-- Archiving configuration
+-- If mod_mam is enabled, Prosody will store a copy of every message. This
+-- is used to synchronize conversations between multiple clients, even if
+-- they are offline. This setting controls how long Prosody will keep
+-- messages in the archive before removing them.
+
+archive_expires_after = "{{ prosody_archive_expires_after }}"
```
madmatah:
```suggestion
-- Archiving configuration
-- If mod_mam is enabled, Prosody will store a copy of every message. This
-- is used to synchronize conversations between multiple clients, even if
-- they are offline. This setting controls how long Prosody will keep
-- messages in the archive before removing them.
archive_expires_after = "{{ prosody_archive_expires_after }}"
```

apps/prosody/templates/services/app/deploy.yml.j2
```diff
@@ -0,0 +1,74 @@
...
+metadata:
+  labels:
+    app: prosody
+    service: app
+    version: "{{ prosody_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+  name: "prosody-app"
+  namespace: "{{ namespace_name }}"
+spec:
+  replicas: {{ prosody_replicas }}
```
madmatah: Does prosody work with multiple replicas?
If not, I suggest to hardcode `replicas: 1`, remove the `affinity:` section and the `prosody_replicas` variable.

apps/prosody/templates/services/app/deploy.yml.j2
```diff
@@ -0,0 +1,74 @@
...
+{% if image_pull_secret_name is not none %}
+      imagePullSecrets:
+        - name: "{{ image_pull_secret_name }}"
+{% endif %}
+      containers:
+        - name: prosody
+          image: "{{ prosody_image_name }}:{{ prosody_image_tag }}"
+          imagePullPolicy: Always
+          resources: {{ prosody_app_resources }}
+          volumeMounts:
```
madmatah, line 35: Since the default value for `prosody_storage` is `internal`, we might want a persistent volume for `/var/lib/prosody` ?

madmatah (commented):
Nice to see new arnold apps :star_struck: 

I noticed a typo in the commit message body : «their service name» instead of «there».

apps/prosody/templates/services/app/deploy.yml.j2
```diff
@@ -0,0 +1,74 @@
...
+                      operator: In
+                      values:
+                        - "prosody-app"
+                topologyKey: kubernetes.io/hostname
+{% set image_pull_secret_name = prosody_image_pull_secret_name | default(none) or default_image_pull_secret_name %}
+{% if image_pull_secret_name is not none %}
+      imagePullSecrets:
+        - name: "{{ image_pull_secret_name }}"
+{% endif %}
+      containers:
```
madmatah, line 30: No `securityContext` is defined for this deployment, is it intentional?

apps/prosody/templates/services/app/deploy.yml.j2
lunika: I have to test this.

apps/prosody/templates/services/app/deploy.yml.j2
lunika: Yes and use it only if the storage is not `sql`

apps/prosody/templates/services/app/deploy.yml.j2
lunika: Yes, I want to use the `prosody` user defined in the image.

apps/prosody/templates/services/app/configs/prosody.cfg.lua.j2
lunika: Yes

apps/prosody/templates/services/app/deploy.yml.j2
lunika: It doesn't work, I will apply your suggestion.

apps/prosody/templates/volumes/storage.yml.j2
```diff
@@ -0,0 +1,18 @@
...
+    app: prosody
+    version: "{{ prosody_image_tag }}"
+    deployment_stamp: "{{ deployment_stamp }}"
+spec:
+  accessModes:
+  - ReadWriteMany
+  resources:
+    requests:
+      storage: {{ prosody_storage_volume_size }}
+  storageClassName: "{{ default_storage_class_rwx }}"
```
madmatah:
```suggestion
  - ReadWriteOnce
  resources:
    requests:
      storage: {{ prosody_storage_volume_size }}
  storageClassName: "{{ default_storage_class_rwo }}"
```

Since there is only one instance of prosody at the same time (replicas:1 + non blue green app), I would suggest to use a `ReadWriteOnce` persistent volume.

apps/prosody/vars/all/main.yml
```diff
@@ -0,0 +1,95 @@
...
+prosody_client_connection_port: 5222
+prosody_nodeport_client_connection_port: 30007
+prosody_http_port: 5280
+prosody_external_components_port: 5347
+prosody_nodeport_external_components_port: 30008
+prosody_secret_name: "prosody-{{ prosody_vault_checksum | default('undefined_prosody_vault_checksum') }}"
+
+# -- Volumes
+# The storage volume is used only when the prosody storage is internal.
+prosody_storage_volume_size: 2Gi
```
madmatah: It seems to be huge to store plain text files :sweat_smile:

madmatah (approved):
LGTM! I forgot one minor feedback on the default volume size.

apps/prosody/vars/all/main.yml
lunika: Indeed 🙂

PR #660, opened 2021-04-28
⚡️(scripts) add docker registry to k3d local cluster

jmaupetit:
## Purpose

When using arnold in a CI context for third-party apps deployable as trays, we often have to build a Docker image of the current source code state and test this code in a k8s context. To do so, we need a local registry to push our newly built image that will be deployed.

## Proposal

- [x] add _ad hoc_ registry creation option while creating the k3d cluster

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,6 +185,32 @@ jobs:
               "${ARNOLD_IMAGE}" \
               ansible --version
 
+  test-k3d-cluster-registry:
+    <<: *defaults
+
+    steps:
+      - checkout
+      - *attach_workspace
+      - *ci_env
+      - *install_k8s_requirements
+
+      - run:
+          name: Test the k3d docker registry
+          command: |
+            # Create the cluster with the registry
+            MINIMUM_AVAILABLE_RWX_VOLUME=15 \
+            K3D_BIND_HOST_PORT_HTTP=80 \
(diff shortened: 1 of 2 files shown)


madmatah (commented):
According to the [k3d documentation](https://k3d.io/usage/guides/registries/#create-a-dedicated-registry-together-with-your-cluster):

> the port, which the registry is listening on will be mapped to a random port on your host system

Is it usable and practical as it is?

jmaupetit:
As stated in the documentation:

> Check the k3d command output or `docker ps -f name=k3d-mycluster-registry` to find the exposed port (let’s use 12345 here)

I wanted to try this.

madmatah:
Okay, I don't find it very practical to be used in scripts like this.

WDYT about making it optional, with a `ENABLE_REGISTRY` environment variable?

jmaupetit:
An alternative implementation would be to (optionally) create a dedicated registry with a fixed port [1]. WDYT?

[1] https://k3d.io/usage/guides/registries/#create-a-customized-k3d-managed-registry

madmatah:
> An alternative implementation would be to (optionally) create a dedicated registry with a fixed port [1]. WDYT?
> 
> [1] https://k3d.io/usage/guides/registries/#create-a-customized-k3d-managed-registry

Personally I prefer this approach :+1:

jmaupetit:
Let's go for it then!

madmatah (approved):

PR #665, opened 2021-06-11
Migrate nextcloud to k8s

sampaccoud:
## Purpose

We want to migrate the `nextcloud` app to kubernetes as well.

## Proposal

I modified the object definitions until the app deployed from my laptop.


lunika (approved):

jmaupetit (approved):

apps/nextcloud/templates/services/app/deploy.yml.j2
```diff
@@ -75,4 +79,8 @@ spec:
           defaultMode: 420
           name: nextcloud-app-{{ deployment_stamp }}
       - name: nextcloud-config
-        emptyDir: {}  # volume that lives as long as the pod lives      
+        # volume that lives as long as the pod lives
+        emptyDir: {}
+      securityContext:
+        runAsUser: {{ container_uid }}
```
madmatah, line 85: I don't know the reason why `nexcloud_user_id` was used before.
Won't that be a problem to use another UID to run the container @lunika?

madmatah (approved):
Just 1 point to clarify, but otherwise LGTM!

apps/nextcloud/templates/services/app/deploy.yml.j2
sampaccoud: The `nextcloud` Docker image is built for a specific user id.
I changed the parameter name to match what we now do on other apps and set it to the value that was previously set for `container_uid`

madmatah: Ok. Maybe you can rename the variable  in `apps/nextcloud/vars/all/main.yml` (`nextcloud_user_id` -> `container_uid`) and add a comment to specify that this container is built for a specific user id?

sampaccoud: @madmatah I didn't rename it but removed it as it is already defined in `groups_vars/all/main.yml` for all applications. Can you check the comment? https://github.com/openfun/arnold/pull/665/files#diff-95b2154a704cb558d83bc7d7d651e817d7e825263183fa3e70166585e5284c91R34

madmatah: The default value for `container_uid` defined in `groups_var/all/main.yml` is set to `10000`.
I don't think it will work out of the box, since the nextcloud image is built with UID `1000` according to your comment.

That's why I suggest to override the value for this `nextcloud` arnold app.

sampaccoud: no no my comment should read `10000` as I was talking about the default `container_uid` value. The comment asks to build your image with the uid you want... we have our own image build but this is our problem.

sampaccoud: Fixed.

madmatah: Oh, ok, It's clear now!
:ship:

PR #666, opened 2021-06-16
🐛(elasticsearch) fix stateful set pvc creation deactivation

sampaccoud:
## Purpose

PVC can be deactivated for an Elasticsearch stateful set. In this case the volumes should not be created.

After deploying the Ashley which contains an Elasticsearch to our new Kubernetes cluster, we noticed that the PVC were created.

## Proposal

Add a condition to skip creating the PVC in case the feature is disabled.


madmatah (approved):

sampaccoud:
https://user-images.githubusercontent.com/1427165/122232233-5e3a0900-cebb-11eb-8780-f76ce139bc05.mp4

PR #670, opened 2021-07-28
✨(prosody) allow to use websocket connection

lunika:
## Purpose

We can configure prosody to use websocket connection and/or bosh
connection. New variable are instroduce the connection to use and how to
configure them.
Also, the read_timeout connection setting can be modified.

## Proposal

- [x] allow to use websocket connection


sampaccoud (approved):
Don't forget the CHANGELOG otherwise GTM

lunika:
I didn't update it because prosody is in the `ADDED` section.

PR #672, opened 2021-08-10
📝(project) update documentation using the CLI

madmatah:
## Purpose

_Note: This PR is based on @jmaupetit 's work on #582. It has been updated to use Kubernetes instead of OKD._

Now that we have a CLI to use arnold, our documentation needed a solid revamp to reflect recent changes.

The documentation no longer focuses on edxapp but the tool itself, with a clear distinction between tool users and developers.


README.md
```diff
@@ -2,178 +2,230 @@
...
-
-## Quick start
-
-**Disclaimer**: this quick start guide has been written with development /
-testing purpose in mind. If you are looking for more insights on how to use it
-in production, please refer to our [documentation](./docs/index.md).
+- [Kubectl](https://kubernetes.io/docs/tasks/tools/):
+  This CLI is used to communicate with the running Kubernetes instance you
+  will use.
+- [k3d](https://k3d.io/): This tool is used to setup and run a lightweight
```
sampaccoud: this is probably optional no? I think the most common case is someone who already has a cluster with a cloud provider and wants to deploy to it so we should not bother them with the complexity of running k8s locally?

If you agree, it means that the bootstrap instructions below should instruct to connect to one's existing k8s cluster or propose to start a k3d as an option (with a link to another doc for that?)

madmatah: I agree with these points:
- `k3d` is an optional dependency
- we need instructions on how to connect to an existing Kubernetes cluster

Since this last part might be complicated (due to the service account requirement), I would be tempted to stick to using k3d in the bootstrap instructions since it's easier like this. But maybe i'm wrong and the real use case is to directly connect to a real kubernetes cluster. Your opinion will be appreciated after reading this.

sampaccoud: ok let's keep it like this for the moment!

sampaccoud: So for the moment I moved k3d to optional dependencies with a mention: https://github.com/openfun/arnold/pull/672/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R25-R28

sampaccoud: Then I removed all outdated documentation files. @madmatah are you ok with that?

madmatah: @sampaccoud Thanks! yes I'm ok with that

sampaccoud (approved):

PR #684, opened 2021-09-16
Allow to configure nginx read and send timeout in prosody application

lunika:
## Purpose

We need to configure the nginx read and send proxy timeout. We decided
to set the same value to the ingress controller.

## Proposal

- [x] enable mod_smacks when websocket is used
- [x] allow to configure nginx reand and send timeout

### DIFF
--- a/apps/prosody/templates/services/nginx/configs/prosody.conf.j2
+++ b/apps/prosody/templates/services/nginx/configs/prosody.conf.j2
@@ -37,7 +37,8 @@ server {
     proxy_set_header Host $host;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
-    proxy_read_timeout 86400;
+    proxy_read_timeout {{ prosody_nginx_proxy_read_timeout }};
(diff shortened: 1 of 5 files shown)


apps/prosody/templates/services/nginx/configs/prosody.conf.j2
```diff
@@ -37,7 +37,8 @@ server {
     proxy_set_header Host $host;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
-    proxy_read_timeout 86400;
+    proxy_read_timeout {{ prosody_nginx_proxy_read_tiemout }};
+    proxy_send_timeout {{ prosody_nginx_proxy_send_tiemout }};
```
madmatah:
```suggestion
    proxy_read_timeout {{ prosody_nginx_proxy_read_timeout }};
    proxy_send_timeout {{ prosody_nginx_proxy_send_timeout }};
```

(+ other occurences in ingress.yml.j2 and main.yml)

lunika: It's already fixed, I saw the typo when I opened the PR... shame on me.

madmatah (approved):

PR #688, opened 2021-09-28
⬆️ (project) upgrade to python 3.9

jmaupetit:
## Purpose

Recent Ansible releases require to upgrade base python release to 3.8+.

## Proposal

- [x] upgrade base python image to 3.9
- [x] unpin outdated secondary dependencies

PR #692, opened 2021-09-29
Add RBAC built-in support

jmaupetit:
## Purpose

Some apps may require additionnal permissions to run. Those permissions should be described in RBAC objects [1].

[1] https://kubernetes.io/docs/reference/access-authn-authz/rbac/

## Proposal

Arnold will now create every RBAC object whose definition file respects the following (glob) naming pattern: `rbac*.yml.j2` in the application service templates directory.

Depends on #688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Implemented support for RBAC objects
+
 ### Changed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,9 +8,14 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Implemented support for rbac objects
```
lunika:
```suggestion
- Implemented support for RBAC objects
```

lunika (approved):

PR #696, opened 2021-10-19
Allow adding alternate routes to ashley

sampaccoud:
## Purpose

Some browsers have started blocking third party cookies from within an iframe. As a result Ashley is broken is broken when integrate in an LMS page.

## Proposal

The solution is to host Ashley on a domain that is a sibling of the domain on which the LMS itself is hosted. We can do that by allowing to declare alternative routes for the ashley app.

### Fixed

- Allow declaring alternative routes for Ashley as a workaround third party cookies being blocked from iframes by some browsers.


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+## [6.5.1] - 2021-10-19
+
+### Fixed
+
+- Allow declaring alternative routes for Ashley as a workaround to third party
+  cookies being blocked from iframes by some browsers
```
lunika:
```suggestion
  cookies being blocked by some browsers
```
They are all blocked, it is not related to an iframe

apps/ashley/templates/services/app/_env.yml.j2
```diff
@@ -2,7 +2,7 @@ DB_ENGINE: "{{ ashley_database_engine }}"
 DB_HOST: "ashley-{{ ashley_database_host }}-{{ deployment_stamp }}"
 DB_NAME: "{{ ashley_database_name }}"
 DB_PORT: "{{ ashley_database_port }}"
-DJANGO_ALLOWED_HOSTS: "{{ ashley_host | blue_green_hosts }}"
+DJANGO_ALLOWED_HOSTS: "{% for host in (ashley_consumer_hosts + [ashley_host]) %}{% if loop.index > 1 %}, {% endif %}{{ host | blue_green_hosts }}{% endfor %}"
```
lunika, line 5: so all the domains will be blue/green ? Is it needed ?

jmaupetit (approved):

apps/ashley/templates/services/app/_env.yml.j2
sampaccoud: We could do without but we don't want to have to remember which one has the blue/green?

CHANGELOG.md
sampaccoud: fixed :pray:

apps/ashley/templates/services/app/_env.yml.j2
lunika: ok :+1:

lunika (approved):

PR #710, opened 2022-01-04
Upgrade elasticsearch and kibana to 7.16.2

jmaupetit:
## Purpose

With the recent log4j CVE, we've considered upgrading default elasticsearch suite releases to the most recent patched version.

## Proposal

- [x] Upgrade `elasticsearch` to `7.16.2`
- [x] Upgrade `kibana` to `7.16.2`


.circleci/config.yml
```diff
@@ -619,8 +619,6 @@ jobs:
             kubectl -n ci-eugene expose pod elasticsearch --port 9200 --name=elasticsearch
             # Deploy Kibana
             bin/arnold -c eugene -e ci -a kibana deploy
-            # Test service deployed with the next route
-            bin/ci route -s kibana -c green -p "/api/status" -P next
```
lunika, line 623: This endpoint does not exists any more?

jmaupetit: No it does, but the configuration is not blue-green compatible (see the `server.publicBaseUrl` in the application ConfigMap). We only test the deployment after having switched it.

lunika (approved):

PR #713, opened 2022-01-10
✨(marsha) configure nginx to handle websocket

lunika:
## Purpose

Marsha supports websocket. We need to configure Nginx to manage this
connection. All websockets begin with `/ws/`, a new location block is
created responsible to deal with the Connection and Upgrade header

## Proposal

- [x] configure nginx to manage websocket

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Marsha: configure nginx to handle websockets in a dedicated location
+
 ## [6.9.0] - 2022-01-17
 
 ### Changed
--- a/apps/marsha/templates/services/nginx/configs/marsha.conf.j2
+++ b/apps/marsha/templates/services/nginx/configs/marsha.conf.j2
@@ -61,6 +61,19 @@ server {
     try_files $uri @proxy_to_marsha_app;
   }
 
+  location /ws/ {
+    proxy_pass http://marsha-backend;
+
+    proxy_http_version 1.1;
+    proxy_set_header Connection "Upgrade";
+    proxy_set_header Upgrade $http_upgrade;
+
+    proxy_buffering off;
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+  }
+
   location /xapi {
 {% if http_basic_auth_enabled and bypass_htaccess %}
     if ($http_x_forwarded_for !~ ^({{ marsha_nginx_bypass_htaccess_ip_whitelist | join("|") }})) {


apps/marsha/vars/all/main.yml
```diff
@@ -15,6 +15,8 @@ marsha_nginx_status_endpoint: "/__status__"
 marsha_nginx_admin_ip_whitelist: []
 marsha_nginx_bypass_htaccess_ip_whitelist: []
 marsha_nginx_static_cache_expires: "1M"
+marsha_nginx_websocket_proxy_read_timeout: 86400
+marsha_nginx_websocket_proxy_send_timeout: 86400
```
jmaupetit: Can you add a comment explaining why a 24h read/send timeout?

apps/marsha/templates/services/nginx/ingress.yml.j2
```diff
@@ -14,6 +14,8 @@ metadata:
 {% if prefix in acme_enabled_route_prefix %}
     cert-manager.io/issuer: "{{ acme_issuer_name }}"
 {% endif %}
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "{{ marsha_nginx_websocket_proxy_read_timeout }}"
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "{{ marsha_nginx_websocket_proxy_send_timeout }}"
```
jmaupetit: This configuration will impact all connections/locations received by nginx, No side effect is expected?

apps/marsha/templates/services/nginx/configs/marsha.conf.j2
```diff
@@ -61,6 +61,22 @@ server {
...
+
+    proxy_buffering off;
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+
+    proxy_read_timeout {{ marsha_nginx_websocket_proxy_read_timeout }};
+    proxy_send_timeout {{ marsha_nginx_websocket_proxy_send_timeout }};
+  }
+
```
jmaupetit, line 76: IMO you don't need a new location if you allow to upgrade connections for the `/` location.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Marsha: configure nginx to manage websockets
```
```suggestion
- Marsha: configure nginx to handle websockets
```

apps/marsha/templates/services/nginx/configs/marsha.conf.j2
lunika: This new location is needed to isolate websocket connection from others. For a classic http connection we don't want to allow a so big timeout. Only for websocket. So we must add this new location dedicated to websocket.

apps/marsha/templates/services/nginx/ingress.yml.j2
lunika: By the ingress. And after in the nginx deploy with marsha we can reconfigure a lower timeout if needed. The first closing the connection will close the connection for all other proxies.

apps/marsha/vars/all/main.yml
lunika: Yes I can

lunika: After tests, keeping the default value still work. The ping/pong implementation keeps alive the connection while the client is connected. I will remove all timeout but I keep the `location` block in the nginx configuration to not mix webstocket and http requests.

apps/marsha/templates/services/nginx/configs/marsha.conf.j2
lunika: One other argument to keep this location block is `http/2`. The UvicornWorker is able to handle http2 and I see that browsers compatible are using this protocol

lunika:
@jmaupetit ready for a new review

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Marsha: configure nginx to handle websockets
```
```suggestion
- Marsha: configure nginx to handle websockets in a dedicated location
```

jmaupetit (approved):

PR #719, opened 2022-01-17
🔖(minor) bump release to 6.9.0

jmaupetit:
### Changed

- Upgrade `ansible` to `5.2.0`

### Fixed

- Kibana: fix configuration to load balance traffic upon replicas


lunika:
Bumping the minor version just because a dependency has been updated, sorry but I disagree.

jmaupetit:
:thinking: 

1. Ansible is the core dependency of the whole thing,
2. Patch releases are supposed to be bug fixes (see the [semver spec](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api))

lunika:
> Ansible is the core dependency of the whole thing,

The core dependency is docker, not Ansible. The whole application is run in a container so there is no effect when you change a dependency.
Ansible is never mention as a requirement in the Readme for example.

> Patch releases are supposed to be bug fixes 

This is why I think it should be the version `6.8.2` and not `6.9.0`

jmaupetit:
Docker is a way to ease running Ansible. Saying so is similar to saying that Marsha is not a Django project!

I don't get your point: upgrading Ansible is not a bug fix. :thinking:

lunika:
It's not a new feature either.

lunika (approved):

PR #726, opened 2022-02-15
🔧(marsha) add env variable DJANGO_CSRF_TRUSTED_ORIGINS in deployment 

lunika:
## Purpose

We have to configure the CSRF_TRUSTED_ORIGINS django settings with all
blue/green domains. But they must be prefixed with the scheme to be
valid.

## Proposal

- [x] add env variable DJANGO_CSRF_TRUSTED_ORIGINS in deployment 
- [x] create release 6.11.0


apps/marsha/templates/services/app/_deploy_base.yml.j2
```diff
@@ -81,6 +81,8 @@ spec:
               value: "{{ marsha_postgresql_port }}"
             - name: DJANGO_ALLOWED_HOSTS
               value: "{{ marsha_host | blue_green_hosts }}"
+            - name: DJANGO_CSRF_TRUSTED_ORIGINS
+              value: "{{ ['https://']| product(marsha_host | blue_green_hosts | split(',')) | map('join') | join(',') }}"
```
sampaccoud: a bit too smart :sweat_smile:  maybe split in two line by setting the first step in a variable?

sampaccoud (approved):

apps/marsha/templates/services/app/_deploy_base.yml.j2
lunika: It looks like Bash. You don't like Bash ?

jmaupetit: Maybe the following suggestion is less complex to figure out?

```suggestion
              value: "{{ marsha_host | blue_green_hosts | split(',') | map('regex_replace', '^(.*)$', 'https://\\1') | join(',') }}"
```

PR #736, opened 2022-04-25
✨(project) allow to use the `deployment_stamp` variable in project variables

jmaupetit:
## Purpose

We often need to configure applications given a deployment stack object name that depends on the deployment_stamp value.

## Proposal

Set the `deployment_stamp` variable before setting project variable to be able to access its value in Arnold's applications configuration.

Quick win. Huge improvement 🎉


sampaccoud (approved):

sampaccoud:
I'm curious what use case do you have for this?

PR #750, opened 2022-08-03
✨(apps) define ingress class name to use on each ingress

lunika:
## Purpose

A default ingress class name is not necessarily defined in a k8s cluster. We can specify which one to use by defining the `ingressClassName` in the ingress spec. The ingress can be defined globally and/or at the application level.

## Proposal

- [x]  define ingress class name to use on each ingress

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Define ingress class name to use on each ingress
+
 ### Changed
 
 - Upgrade `ansible` to `6.1.0`
--- a/group_vars/all/main.yml
+++ b/group_vars/all/main.yml
@@ -181,3 +181,8 @@ container_gid: 0
 # directly.
 default_storage_class_rwx: "manual"
 default_storage_class_rwo: "local-path"
+
+# Ingress class name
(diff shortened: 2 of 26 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- define ingress class name to use on each ingress
```
```suggestion
- Define ingress class name to use on each ingress
```

group_vars/all/main.yml
```diff
@@ -181,3 +181,8 @@ container_gid: 0
 # directly.
 default_storage_class_rwx: "manual"
 default_storage_class_rwo: "local-path"
+
+# Ingress class name
+# A default ingress class name is not necessarily defined in a k8s cluster.
+# We can specify which one to use by defining the ingressClassName in the Ingress spec
+default_ingress_class_name: "nginx"
```
jmaupetit: A carriage return is missing:

```suggestion
default_ingress_class_name: "nginx"

```

jmaupetit (commented):
I only have minor comments, thank you for this!



:warning: There are minor typos In your commit message, here is my proposal:

```
A default ingress class name is not necessarily defined in a k8s
cluster. We can specify which one to use by defining the
ingressClassName in the ingress spec. The ingress can be defined
globally and/or at the application level.
```

jmaupetit (approved):

PR #758, opened 2022-11-08
✨(marsha) add new cronjob managing clean_aws_elemental_stack command

lunika:
## Purpose

There is a new management command clean_aws_elemental_stack that should be managed by a cronjob in our k8s.

## Proposal

- [x] Add new cronjob managing clean_aws_elemental_stack command

### DIFF
--- a/apps/marsha/vars/all/main.yml
+++ b/apps/marsha/vars/all/main.yml
@@ -46,6 +46,7 @@ marsha_check_live_state_cronjob_schedule: "*/20 * * * *"
 marsha_check_idle_state_cronjob_schedule: "0 4 * * *"
 marsha_clean_mediapackages_cronjob_schedule: "0 5 * * *"
 marsha_clean_medialive_dev_stack_cronjob_schedule: "*/5 * * * *"
+marsha_clean_aws_elemental_stack_cronjob_schedule: "0 */2 * * *"
 
 # -- volumes
 marsha_media_volume_size: 2Gi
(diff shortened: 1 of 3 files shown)


jmaupetit (commented):
Given the growing number of almost identical cronjob objects you create in this app, what about creating a cronjob list instead and moving executed commands / scheduling in a configuration variable as it's achieved in Ralph's Tray:

- the template: https://github.com/openfun/ralph/blob/master/src/tray/templates/services/app/cronjob_pipeline.yml.j2
- the configuration: https://github.com/openfun/ralph/blob/master/src/tray/vars/all/main.yml#L16-L29

apps/marsha/vars/all/main.yml
```diff
@@ -75,3 +65,32 @@ marsha_postgresql_resources:
   requests:
     cpu: 10m
     memory: 100Mi
+
+# -- Cronjobs
+marsha_deploy_clean_medialive_dev_stack_cronjob: false
```
jmaupetit: Is this variable used elsewhere?

lunika: In the project no, but this is the only redefine by environment

apps/marsha/vars/all/main.yml
```diff
@@ -75,3 +65,32 @@ marsha_postgresql_resources:
...
+  - name: "clean-medialive-dev-stack"
+    schedule: "*/5 * * * *"
+    command: "clean_medialive_dev_stack"
+    resources: "{{ app_resources }}"
+    enabled: {{ marsha_deploy_clean_medialive_dev_stack_cronjob }}
+  - name: "clean-mediapackages"
+    schedule: "0 5 * * *"
+    command: "clean_mediapackages"
+    resources: "{{ app_resources }}"
+    enabled: true
```
jmaupetit: What about defining those in the deployment project and define an empty list as the default `marsha_cronjobs` value?

lunika: Because I don't want to copy past this everywhere. This is the limit of this refacto IMO

lunika: And this list is needed to have a complete marsha environment working.

lunika:
@jmaupetit I removed my last commit, I have to investigate why I can't override the cronjob list then in matsuo. 

I made a backup of the commit in an other branch and will wotk on it later. 
Can you finish the review please considering I will do your suggestion later ?
Thanks

jmaupetit (approved):
Maybe declare a new issue for the WIP improvement!

PR #766, opened 2022-12-30
✨(marsha) manage service worker path in nginx

lunika:
## Purpose

The service work we are using must be served at the root of the application. But we are managing it as a static file and is saved we the other static file managed by django. We need to add a rule in the nginx conf to access it from the root.

## Proposal

- [x] manage service worker path in nginx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- Marsha: Add nginx rule to manage service worker
+
 ### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
```
sampaccoud, line 11: in commit message: 
s/service work/service worker
s/and is saved we the other static file/and IT is saved WITH the other static fileS

sampaccoud (approved):

PR #777, opened 2023-06-21
🔧(kibana) remove server UUID generation

jmaupetit:
## Purpose

Recent Kibana releases auto-generates a uuid to uniquely identify a server instance. It is no longer required to generate and set it manually.

## Proposal

- [x] remove server UUID generation mechanism


quitterie-lcs (approved):
LGTM

PR #779, opened 2023-07-31
Allow having multiple ingresses for an app 

lunika:
## Purpose

In a single app we want to manage multiple ingresses targeting the same
service. To do this, we want to generate multiple yaml documents in once
and we have to switch on the the from_yaml_all filter in ansible.

## Proposal

- [x] Allow having multiple ingresses for an app 
- [x] bump release to 6.19.0


apps/hello/templates/services/app/ingress.yml.j2
```diff
@@ -1,3 +1,5 @@
+{% for hello_host in hello_hosts %}
```
claudusd, line 1: You create an ingress for each ``hello_hosts`` but the ingress name didn't change _(see line 6)
https://github.com/openfun/arnold/pull/779/files#diff-fa38320aafeaf3e81ee9907e54e8833a78f10f5a4ffe0511b2b5dfc05e14bbd2L4
If you are in different namespace this could not be a problem but in the same namespace this will be override.

I think you should add rules in your ingress specification for each host.

``` yml.j2
spec:
  ingressClassName: "{{ hello_ingress_class_name }}"
  rules:
  {% for hello_host in hello_hosts %}
  - host: "{{ hello_host | blue_green_host(prefix) }}"
    http:
      paths:
      - backend:
        service:
          name: "hello-app-{{ prefix }}"
          port:
            number: {{ hello_app_port }}
        path: /
        pathType: Prefix
  {% endfor %}
  tls:
  {% for hello_host in hello_hosts %}
  - hosts:
    - "{{ hello_host | blue_green_host(prefix) }}"
    secretName: "hello-app-tls-{{ prefix }}-{{ acme_env }}"
  {% endfor %}
```

claudusd (changes requested):

apps/hello/templates/services/app/ingress.yml.j2
lunika: Nice idea. With this I don't need anymore this PR because I can do it with only one yaml document. Thanks

PR #780, opened 2023-08-22
✨(apps) use variable to select PVC to use

claudusd:
## Purpose

The persistent volume name are hard-coded in the resources definition. This PR allow to override this name.

## Proposal

Add a specific variable for each PVC to override the name. If the variable isn't defined we use the original name to avoid breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Add variables to override persistent volume names
+
 ## [6.18.0] - 2023-07-03
 
 ### Changed
--- a/apps/edxapp/vars/all/main.yml
+++ b/apps/edxapp/vars/all/main.yml
@@ -26,6 +26,16 @@ edxapp_i18n_languages:
   - fr
 edxapp_tx_secret_name: "edxapp-tx-{{ edxapp_vault_checksum | default('undefined_edxapp_vault_checksum') }}"
 
+# -- pvc
+# Persitent volume name use for media
+edxapp_pvc_media_name: "edxapp-pvc-media"
+edxapp_pvc_data_name: "edxapp-pvc-data"
+edxapp_pvc_export_name: "edxapp-pvc-export"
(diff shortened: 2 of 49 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- Add variables to override persistent volume name
```
```suggestion
- Add variables to override persistent volume names
```

apps/edxapp/vars/all/main.yml
```diff
@@ -26,6 +26,9 @@ edxapp_i18n_languages:
   - fr
 edxapp_tx_secret_name: "edxapp-tx-{{ edxapp_vault_checksum | default('undefined_edxapp_vault_checksum') }}"
 
+# -- pvc name
+edxapp_pvc_media_name: "edxapp-pvc-media"
```
jmaupetit, line 31:
```suggestion
edxapp_v_media_pvc_name: "edxapp-pvc-media"
```

jmaupetit (commented):
Instead of providing defaults in templates, shouldn't we define vars explicitly in `{{ app }}/vars/all/main.yml` so that every variable is declared / auto-documented somewhere (a la helm)?

apps/edxapp/vars/all/main.yml
claudusd: Not in this case, there are 2 variables.

First one ``edxapp_pvc_media_name`` is the name of the PV create here https://github.com/openfun/arnold/blob/681dba5705439c4ffb8ab8b14c3d552c94ff5c47/apps/edxapp/templates/volumes/media.yml.j2#L4
It's the only volume with a variable to specified the PVC name. I keep it to avoid breaking change.

The second ``edxapp_v_media_pvc_name`` is to select the PVC to use for the edxapp's media volume.

claudusd:
@jmaupetit, I replace all default with the variable initialization in app default vars file.

lunika (approved):
your commits can be squash IMO

lunika:
You also missed to variabilize the PVC in each volumes file. This should also be made to be consistent.

lunika (approved):

PR #782, opened 2023-09-07
🛂(cli) use github token to avoid rate limiting

claudusd:
## Purpose

If we execute a lot of Arnold command, this will reach the Github API rate limiting. 
When we reach the limit Arnold will fail because it can parse the Github response and stop with this error
```
jq: error (at <stdin>:1): Cannot index object with number
```

## Proposal

Use the environment variable ``GITHUB_TOKEN`` to authenticate to the Github API and increase the API rate limiting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
-### Changed
+### Added
+
+- Use Gitlab token when call the Github's API
 
 ## [6.19.0] - 2023-09-01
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,9 @@ Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
-### Changed
+### Added
+
+- Use Gitlab token when call the Github's API
```
jmaupetit, line 13:
```suggestion
- Use GitHub token to call Github's API when the `GITHUB_TOKEN` environment variable is defined
```

bin/arnold
```diff
@@ -254,9 +254,14 @@ function _get_log_level_name() {
 
 # Get the latest Arnold release by querying Github's API
 function _get_latest_release() {
+  local auth_header=()
```
jmaupetit, line 257: I prefer `declare` which is more explicit as it can cast vars:

```suggestion
  declare -a auth_header
```

PR #797, opened 2024-10-01
Deploy multiple yaml document

lunika:
## Purpose

We want to allow having multiple documents in a yaml file. For this
instead of using the filter from_yaml we have to use from_yaml_all.

## Proposal

- [x] allow to load multiple documents in yaml templates


jbpenrath (commented):
No need here ? 
https://github.com/openfun/arnold/blob/a3572a9af22beea87bee0ffa00c4063702e9812e/tasks/create_redirection_ingresses.yml#L12

lunika:
> No need here ?
> 
> https://github.com/openfun/arnold/blob/a3572a9af22beea87bee0ffa00c4063702e9812e/tasks/create_redirection_ingresses.yml#L12

I choose to not modify it here because it is an app managed by arnold directly and we control it. It's not the case for community apps.

jbpenrath (approved):

claudusd (approved):

